PR #10, opened 2022-04-21
Add special handling for the Copy trait

voidc:
This finishes the implementation from yesterday's walkthrough, adding a missing where clause and fixing the #:precondition.
Now, both tests in `decl/test/copy.rkt` pass.
I hope it's okay to open a PR for this. :smile:

### DIFF
--- a/src/decl/test/copy.rkt
+++ b/src/decl/test/copy.rkt
@@ -1,0 +1,83 @@
+#lang racket
+(require redex/reduction-semantics
+         "../decl-to-clause.rkt"
+         "../decl-ok.rkt"
+         "../grammar.rkt"
+         "../prove.rkt"
+         "../../util.rkt")
+
+;; Test the special rules for impls of the Copy trait.
+
+(module+ test
+  (redex-let*
+   formality-decl
+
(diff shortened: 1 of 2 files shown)


src/decl/test/copy.rkt
```diff
@@ -0,0 +1,83 @@
...
+     Env (term (env-for-crate-decl CrateDecl)))
+    )
+
+   (traced '()
+           (decl:test-cannot-prove
+            Env
+            (crate-ok-goal (CrateDecl) CrateDecl)
+            ))
+   )
+  )
```
bjorn3: nit: missing trailing newline

ltentrup:
According to https://doc.rust-lang.org/reference/special-types-and-traits.html#copy one also has to check that `Drop` is not implemented for a type that implements `Copy`. As far as I can tell this can't be stated currently as there is no logical negation nor a predicate that asserts that a trait is not implemented.

nikomatsakis:
Ah, interesting. Good catch @ltentrup, let's file an issue about it! We can certainly add negation of this kind -- I would not want it to be part of the *where clauses* though.

nikomatsakis (approved):

PR #11, opened 2022-04-25
Fix doc comment

pierwill:


pierwill:
Actually, I think maybe this could go either way.

PR #13, opened 2022-04-29
Add typing rules for MIR

voidc:
This adds basic typing rules for MIR.
Currently still work-in-progress.


nikomatsakis:
@voidc needs rebase :)

nikomatsakis:
@voidc let's discuss next steps here!  I think this is going in the right direction but it probably makes sense to map out the plan a bit in advance.

voidc:
@nikomatsakis definitely! I'm currently still on vacation but will continue to work on this next week. We could talk about the next steps during the meeting on Wednesday or before - whatever works for you.

voidc:
I updated the typing rules to match the new grammar. Also I added the `typing-goals` file which contains a translation of the typing rules into goals.

nikomatsakis:
Going to merge this, thanks! I'm doing some local experiments based on this branch to decide on the best overall approach.

PR #15, opened 2022-05-04
Make MIR grammar more consistent with other grammars

voidc:


src/mir/grammar.rkt
```diff
@@ -1,74 +1,65 @@
...
+          (RvalueBinaryOp BinaryOp Operand Operand)
+          (RvalueNullaryOp Ty)
+          )
 
-  (binary-op + - * /)
+  (BinaryOp ::= + - * /)
 
-  (lifetime lifetime-id)
+  (Terminator ::=
+              (TerminatorGoto BasicBlockId)
```
digama0, line 31:
```suggestion
              (goto BasicBlockId)
```
I think the namespacing isn't necessary since these grammar groups are contextual anyway, and keywords are lowercase

voidc: I think the namespacing helps in some cases (e.g. `RvalueUse`, which could be confused with use declarations). So, I kept the prefixes for consistency. I'm not sure whether Goto should be lowercase since the `goto` keyword does not actually appear in Rust surface syntax.

digama0: I believe the capital / lowercase convention is primarily intended to distinguish terminals and nonterminals in redex code. Given the way redex works, I think having such a convention is important, although currently it is not being used uniformly:

* predicates are capital: `WellFormed`, `HasImpl`, `Normalize`
* same for `WhereClause`: `ForAll`, `Implemented`, `Normalize`
* `Ty` constructors are capitalized and sometimes namespaced: `TyRigid`, `TyAlias`, `ForAll` 
* `RigidTy` is a nonterminal, `TyRigid` is a terminal

I wasn't under the impression that all lowercase terminals are supposed to be (derived from) rust surface syntax; there are things like `-outlives-` that don't look anything like rust, and `crate` is not used in the same way as in rust. Personally I would just go for a best effort attempt to look like rust here. CC: @nikomatsakis, how much of this is the intended naming convention and how much is incomplete refactoring of old code with a different convention?

The other reason to avoid namespacing in terminals is that these have to be written out in tests, and we generally want to keep things as short as possible since writing ASTs directly is hard enough as it is. Since these are MIR grammar things, I would suggest we attempt to look like the MIR output produced by rustc where possible, and this uses `goto -> bb1;` for such terminators.

nikomatsakis: Yeah, you've kind of identified the conventions I was shooting for; you've also identified the fact that I've had a certain amount of ambivalence about what is the best naming convention overall.

In general I did try to avoid lowercase words that are not Rust keywords; I adopted `-outlives-` partly because it was not a valid Rust identifier. 

I definitely was not using lowercase as a strict *terminal* identifier, but it kind of evolved a bit by default because I have tried to use Rust keywords for terminals where I can.

Lately I've been wondering about using more symbols, so e.g. we could say `Terminator/Resume` instead of `TerminatorResume`. That said, I think it's probably good to optimize a bit for readaiblity: I'm a bit concerned that the `TerminatorDrop` names etc will make MIR written in PLT Redex form significantly less fun to read. 

I think overall I lean towards: let's use lowercase keywords here, it'll make the MIR nicer to read and I think that's important.

nikomatsakis (changes requested):

voidc:
> I'm going to merge this, but I think we agreed to change the nonterminal convention.

Thanks! I'm currently on vacation but can continue with the discussed renaming next week.

PR #20, opened 2022-05-17
some more terminal naming fixes

digama0:
Some additional cleanup after #18.

* Fixed `TyVar` -> `type` and `TyApply` -> `rigid-ty` in documentation
  * some remaining uses in blog/2022-50-12.mdx but I left it alone since it looks like a historical document
* Renamed `TyUnit` -> `unit-ty`, because even though it is a term definition rather than a primitive terminal constant, it is closer to a constant than a variable (and this also matches the use of other similar metafunctions like `(vec T)`). I thought about removing the `-ty`, but I think this is a non-context-sensitive macro expansion and I wouldn't want other uses of `unit` outside the type grammar to get expanded.
* Some trailing whitespace and missing final newline issues were auto-fixed by my vscode config in the files I touched. I can disable this behavior or roll it back if people care.

### DIFF
--- a/book/src/closer-look-decl.md
+++ b/book/src/closer-look-decl.md
@@ -30,8 +30,8 @@ A trait declaration looks like this:
 Here:
 
 * `TraitId` is the name of the trait
-* `KindedVarIds` is a list of generic parameters like `((TyVar Self) (TyVar T))`.
-  Note that the `Self` parameter is made explicit. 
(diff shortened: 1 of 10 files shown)


nikomatsakis:
thanks!

nikomatsakis:
Thanks! https://github.com/nikomatsakis/a-mir-formality/issues/21 is another cleanup I had in mind along these lines, too.

nikomatsakis (approved):

PR #38, opened 2022-05-22
Get the most basic fn test working

jackh726:
Kind of a small change, but it's a start for me :)

### DIFF
--- a/src/decl/decl-ok.rkt
+++ b/src/decl/decl-ok.rkt
@@ -58,9 +58,15 @@
    ;; We generate the following goal, which specifies that -- assuming the generics
    ;; are well formed and the where-clauses hold -- the field types are well-formed:
    ;;
-   ;;     FIXME
+   ;;     (∀ ((lifetime A) (type T))
+   ;;         (implies ((well-formed (lifetime A))
+   ;;                   (well-formed (type T))
+   ;;                   (well-formed (type (rigid-ty (ref ()) (A (rigid-ty T ()))))))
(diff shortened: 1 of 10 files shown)


nikomatsakis (approved):

src/decl/decl-ok.rkt
```diff
@@ -58,9 +58,15 @@
...
    ;; are well formed and the where-clauses hold -- the field types are well-formed:
    ;;
-   ;;     FIXME
+   ;;     (∀ ((lifetime A) (type T))
+   ;;         (implies ((well-formed (lifetime A))
+   ;;                   (well-formed (type T))
+   ;;                   (well-formed (type (rigid-ty (ref ()) (A (rigid-ty T ()))))))
+   ;;           (well-formed-where-clause-goal (T : Trait_Ord ()))))
+   ;;
+   ;; FIXME: Actually implement that, along with for the other items
```
jackh726, line 67: Thinking more about this. I don't think this is correct. I think we can assume these bounds to be true here and instead we need to prove them at the call site.

nikomatsakis: Well, the way Rust *currently* works, we do require bounds to be WF ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=977f7dce73f98868d16f0208be0f4db2)):

```rust
trait Foo<T: Ord> { }

fn bar<U, T>()
where T: Foo<U>
{}

fn main() {}
```

I kind o agree we shouldn't, but that's implied bounds for ya.

nikomatsakis (approved):

PR #39, opened 2022-05-25
Add basic GATs test

jackh726:
Not sure if this correct?

### DIFF
--- a/src/ty/test/test-gats.rkt
+++ b/src/ty/test/test-gats.rkt
@@ -18,7 +18,7 @@
                  (; Define a trait `Iterable` that is implemented for Vec<T>
                   ∀ ((type T)) (is-implemented (Iterable ((vec T)))))
                  (; normalizes-to `Item<Vec<T>, A>` to `(rigid-ty (ref ()) (A T))`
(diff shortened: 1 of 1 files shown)


src/ty/test/test-gats.rkt
```diff
@@ -0,0 +1,41 @@
...
+  (redex-let*
+   formality-ty
+
+   ((Env (term (env-with-clauses-invariants-and-generics
+                ((; Just ignore well-formed rules, not interesting for testing subtyping
+                  ∀ ((type T)) (well-formed (type T)))
+                 (; Define a trait `Iterable` that is implemented for Vec<T>
+                  ∀ ((type T)) (is-implemented (Iterable ((vec T)))))
+                 (; normalizes-to `Item<Vec<T>, A>` to `(rigid-ty (ref ()) (A T))`
+                  ∀ ((type T) (lifetime A)) (normalizes-to (alias-ty Item ((vec T) A)) (rigid-ty (ref ()) (A T))))
```
nikomatsakis:
```suggestion
                  ∀ ((type T) (lifetime A)) (normalizes-to (alias-ty Item ((vec T) A)) (& A T)))
```

nikomatsakis (commented):
This seems roughly right, but what we really need of course is to have these rules get lowered and produced from trait definitions.

PR #47, opened 2022-05-31
Revise Introducting and sketch philosophy

pierwill:
Some of what I've written here could be controversial. However, I'd like the docs for this project to make a real contribution, moving forward the way we think about langauge maintainance and design, democratizing formal methods in the Rusty spirit that has always tried to empower users. 

This will not be the last revision, not by a long shot. But I hope this moves us forward.

### DIFF
--- a/docs/docs/intro.md
+++ b/docs/docs/intro.md
@@ -7,8 +7,8 @@ sidebar_position: 1
 **A MIR Formality** (or just Formality) is an early-stage experimental project that aims to be
 a complete, authoritative formal model of the [Rust MIR][].
 The project has two goals.
-First, the models of Rust's MIR given by Formality's will serve as a complete model of _safe Rust_.
-including all static checks and operational semantics.
+First, the models of Rust's MIR given by Formality's will serve as a complete model for Rust, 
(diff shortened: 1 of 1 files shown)


docs/docs/intro.md
```diff
@@ -6,6 +6,11 @@ sidebar_position: 1
 
 **A MIR Formality** (or just Formality) is an early-stage experimental project that aims to be
 a complete, authoritative formal model of the [Rust MIR][].
+The project has two goals.
+First, the models of Rust's MIR given by Formality's will serve as a complete model of _safe Rust_.
+including all static checks and operational semantics.
```
nikomatsakis:
```suggestion
First, the models of Rust's MIR given by Formality's will serve as a complete model for Rust, 
including both its type checking rules (what programs are accepted) and semantics (what happens when you run them on an idealized machine).
```

I think we eventually aim to include more than just safe Rust!

nikomatsakis (changes requested):
This looks good! I left one caveat.

nikomatsakis (approved):

PR #48, opened 2022-05-31
add support for statics

### LINKED ISSUE #24: Add support for statics
To support #22 we need to support `static` declarations. We have some support for `const` already, so this would be similar.

Pieces:

* [ ] Add declaration to the decl layer
* [ ] Extend with MIR (depends on #25)
* [ ] Check that type is Send, Sync (depends on #26)

nikomatsakis:
and require that statics are Send

Fixes #24

### DIFF
--- a/src/decl/decl-ok.rkt
+++ b/src/decl/decl-ok.rkt
@@ -148,8 +148,8 @@
         (where-clause->hypothesis WhereClause) ...)
        (&& ((; ... then the trait must be implemented
              well-formed (type Ty))
-            (; ... and the type must be Send
-             is-implemented (rust:Send (Ty)))
+            (; ... and the type must be Sync
+             is-implemented (rust:Sync (Ty)))
             (; ... and the where-clauses must be WF
              well-formed-where-clause-goal CrateDecls WhereClause)
             ...))))
--- a/src/decl/decl-ok.rkt
+++ b/src/decl/decl-ok.rkt
@@ -133,6 +133,28 @@
    (where/error (WhereClause ...) WhereClauses)
    ]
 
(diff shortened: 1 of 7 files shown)


src/decl/decl-ok.rkt
```diff
@@ -133,6 +133,31 @@
...
+   (∀ KindedVarIds
+      (implies
+       (; assuming all generic parameters are WF...
+        (well-formed KindedVarId) ...
+        ; ...where-clauses are satisfied...
+        (where-clause->hypothesis WhereClause) ...)
+       (&& ((; ... then the trait must be implemented
+             well-formed (type Ty))
+            (; ... and the type must be Send
+             is-implemented (rust:Send (Ty)))
```
digama0: Statics don't have to be Send, but they do have to be Sync:
```rust
struct Foo(*const ());
// unsafe impl Send for Foo {}
unsafe impl Sync for Foo {}
static FOO: Foo = Foo(0 as _); // OK
```

src/decl/decl-to-clause.rkt
```diff
@@ -295,6 +295,15 @@
    (() () ())
    ]
 
+  [;; For a named static in the crate C, like the following
+   ;;
+   ;;    static NAMED<T>: Foo<T> where T: Trait;
+   ;;
+   ;; we don't yet any rules.
```
digama0:
```suggestion
   ;; we don't yet enforce any rules.
```

src/decl/decl-ok.rkt
nikomatsakis: Whoops :)

PR #53, opened 2022-06-02
describe x11 setup

nikomatsakis:
cc @jackh726 -- you had some issues around this, right?

### DIFF
--- a/docs/docs/setup.md
+++ b/docs/docs/setup.md
@@ -58,7 +58,7 @@ Gtk initialization failed for display ":0"
 
 then this is indeed the problem.
 
-The problem is that the language server works by running a readless version of racket and GUI functionality is required for some of that, even if it isn't used.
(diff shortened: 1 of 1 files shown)


docs/docs/setup.md
```diff
@@ -40,3 +42,46 @@ The `(traced '() expr)` macro is used to wrap tests throughout the code. The `'(
...
+If you check output and see errors like this:
+
+```
+Unable to init server: Could not connect: Connection refused
+Gtk initialization failed for display ":0"
+```
+
+then this is indeed the problem.
+
+The problem is that the language server works by running a readless version of racket and GUI functionality is required for some of that, even if it isn't used.
```
bjorn3: *headless

jackh726:
There's another workaround that requires adding a setting to vscode. Will have to check later what it is. (I posted it somewhere I think)

PR #67, opened 2022-06-24
"inline" WF goals in impls, make WF a builtin predicate for efficiency

### LINKED ISSUE #61: "self-supporting" well-formedness for associated types
#58 is implementing a lot of the rules for associated types, but there is a bug in the test `src/decl/test/assoc-type-values-wf.rkt`. The test contains a snippet like this...

```rust
trait Iterator { type Item; }
impl<T> Iterator for (T, ) { type Item = MyIter<T>; }
struct MyIter<T: Copy> { }
```

This test is meant to fail, because the impl needs to prove that `MyIter<T>` is well-formed, and to do that, it needs to prove that `T: Copy`.

However, the test passes. It passes because we have a rule that says

```
(is-implemented (Iterator[T])
---------------------
(well-formed (type (< T as Iterator[] > :: Item[])))
```

i.e., `<T as Iterator>::Item` is well-formed if `T: Iterator`.  This is a reasonable rule, but it's meant to be used in *other* code that is working against "some iterator" `T` to determine when it's value to use `T::Item`. 

In this case, we see that we can normalize `<(T,) as Iterator>::Item` to `MyIter<T>` and so we are happy.

This is really a case of "self-supporting impls", similar to what we encountered when trying to permit coinductive impls.

### LINKED ISSUE #62: re-enable "assoc-type-values-meet-trait-bounds.rkt" tests
In #58, the tests in `assoc-type-values-meet-trait-bounds.rkt` are disable. Some of them (particularly, iirc, the 3rd one) send the solver into an infinite loop. The problem is that, when it is trying to prove the type `Once<T>` well-formed, it goes off into an infinite loop. This is because it tries to apply the rule that says `<?T as Iterator>::Item` is WF if `?T: Iterator`, when essentially means "find me a type `?T` that normalizes to `Once<T>`". In so doing, it winds up repeating itself over and over again. This is related but not the same as #61.

nikomatsakis:
This PR changes how we handle well-formedness checking to address #61 and #62. It has two parts:

* In impls, given `type Foo = SomeType`, we no longer try to prove `(well-formed (type SomeType))`. As explained in #61, that is unsound. Instead, we have a function that "inlines" the goals for `SomeType` to be well-formed by one step. (We should probably use this function in some other places...).
* To improve solver efficiency and flexibility, we make "well-formed" a builtin predicate, and add an `ambiguous` goal that causes the solver to reach a "neither true nor false" state. When we try to prove `(well-formed (type ?T))` for some unbound inference variable, we don't go off enumerating all possible well-formed types, we instead mark it as ambiguous, prompting the solver to do reordering. When the type we are proving *is* known, using a built-in predicate allows us to avoid the trap of trying to find out "maybe there is some associated type that normalizes to this!". This makes things *much* faster, fixing #62. 

Fixes #61 
Fixes #62


src/decl/decl-ok/impl-item.rkt
```diff
@@ -78,12 +79,37 @@
...
+   ;
+   ; This other type `SomeOtherType<String>` is not well-formed, so the impl should be invalid.
+   ; But if we just tried to prove `(well-formed (type SomeOtherType<String>))`, the solver would be
+   ; within its rights to combine the facts that ...
+   ;
+   ; * `<SomeType as Iterator>::Item` (as an alias) is well-formed, since `SomeType: Iterator`
+   ; * `<SomeType as Iterator>::Item` normalizes to `SomeOtherType<String>` (from the impl)
+   ; * therefore `SomeOtherType<String>` is WF
+   ;
+   ; Instead we `well-formed-goal-for-ty`, which means we have to prove the
```
voidc:
```suggestion
   ; Instead we use `well-formed-goal-for-ty`, which means we have to prove the
```

src/decl/decl-ok/impl-item.rkt
```diff
@@ -78,12 +79,37 @@
...
+   ; where-clause from `SomeOtherType` definition, so we try to prove `String: Copy` (which fails).
+   ;
+   ; One exception is if we have something like `impl<T> Foo for Blah<T> { type Item = T; }`.
+   ; In that case, the `well-formed-goal-for-ty` will just be `(well-formed (type T))`, which is fine
+   ; since that only provable via a hypothesis (that the one who relies on our impl must prove).
```
voidc:
```suggestion
   ; since that is only provable via a hypothesis (that the one who relies on our impl must prove).
```

src/decl/builtin-predicate/well-formed.rkt
```diff
@@ -0,0 +1,127 @@
...
+   ]
+
+  [(well-formed-goals CrateDecls (rigid-ty (ref _) (Lt Ty)))
+   [(Ty -outlives- Lt)]
+   ]
+
+  ;; ALIAS TYPES
+  ;;
+  ;; The behavior here is a key point of departure from the normal solver, and in fact
+  ;; the entire reason this predicate is builin. If we are verifying that
```
voidc:
```suggestion
  ;; the entire reason this predicate is builtin. If we are verifying that
```

src/logic/grammar.rkt
```diff
@@ -63,6 +74,12 @@
   ;; `Predicate` -- the atomic items that we can prove
   (Predicates ::= (Predicate ...))
 
+  ;; `Predicate/Deboned` -- A *deboned* predicate separates out the "rigid part"
+  ;; (the skeleton) from input/output parameters. To determine whether two
+  ;; predicates ate equal, the skeletons can just be compared for
```
voidc:
```suggestion
  ;; predicates are equal, the skeletons can just be compared for
```

voidc (commented):
found some typos

src/decl/builtin-predicate/well-formed.rkt
```diff
@@ -0,0 +1,127 @@
...
+  ;; variables, though it's not harmful if it is. It will
+  ;; simply return a generic `well-formed` goal that will
+  ;; yield ambiguous when you try to prove it (until that
+  ;; variable is later constrained in some other way).
+  well-formed-goals : CrateDecls Ty -> Goals
+
+  ;; VARIABLES
+  ;;
+  ;; We yield false-goal here -- we don't have any rules to prove
+  ;; random variables well-formed. If they're universal,
```
pierwill, line 64:
```suggestion
  ;; We yield false-goal here -- we don't have any rules to prove that
  ;; random variables are well-formed. If they're universal,
```

Does "random" have a technical meaning in this context? Maybe "arbitrary"?

src/decl/decl-ok/impl-item.rkt
```diff
@@ -78,12 +79,37 @@
...
+   ; leveraging this very impl to prove WF. Imagine this program:
+   ;
+   ; ```
+   ; impl Iterator for SomeType { type Item = SomeOtherType<String>; }
+   ; struct SomeOtherType<T: Copy> { t: T }
+   ; ```
+   ;
+   ; This other type `SomeOtherType<String>` is not well-formed, so the impl should be invalid.
+   ; But if we just tried to prove `(well-formed (type SomeOtherType<String>))`, the solver would be
+   ; within its rights to combine the facts that ...
```
pierwill, line 95:
```suggestion
   ; However, if we tried to prove `(well-formed (type SomeOtherType<String>))` directly, the solver would be
   ; within its rights to combine the facts that ...
```

Maybe something like "the solver could (erroneously) combine the facts" or "the solver might (erroneously) combine the facts"? Would the solver definitely make this mistake given this example, or is it a more general concern that depends on other details as well?
```suggestion
   ; But if we just tried to prove `(well-formed (type SomeOtherType<String>))`, the solver would be
   ; within its rights to combine the facts that
```

PR #74, opened 2022-06-29
make normalization inductive

### LINKED ISSUE #68: normalization can be "unproductive" -- make it inductive?
So the current solver treats all predicates as coinductive in a-mir-formality right now. I think this is (ultimately, RFC required) the right call for trait matching, but for associated type normalization -- at least as current managed -- it is not sound. Consider:

```
trait Foo { type FooItem; }
trait Bar { type BarItem; }

impl Foo for () {
    type FooItem = <() as Bar>::BarItem;
}

impl Bar for () {
    type BarItem = <() as Foo>::FooItem;
}
```

Now imagine that we have a goal like

```
forall<T> { normalizes-to(<() as Foo>::FooItem, T) }
```

This ought not to be provable -- you can't normalize `FooItem` to *any type `T`*. But it is provable if we are naively coinductive, because you get

* `normalizes-to(<() as Foo>::FooItem, T)` if
* `normalizes-to(<() as Bar>::BarItem, T)` if
* `normalizes-to(<() as Foo>::FooItem, T)` -- cycle, ok!

nikomatsakis:
Implemented during the a-mir-formality hacking session, video being posted soon.

Fixes #68


src/decl/builtin-predicate.rkt
```diff
@@ -2,26 +2,28 @@
...
          well-formed-goal-for-ty
          )
 
 (define-metafunction formality-decl
   ;; Part of the "hook" for a formality-decl program:
   ;;
-  ;; Create the clauses for solving a given predicate
-  ;; (right now the predicate is not used).
-  decl:is-builtin-predicate? : Goal -> boolean
+  ;; Extends ty:categorize-goal with a knowledge of which predicates
```
pierwill:
```suggestion
  ;; Extends ty:categorize-goal with information about of which predicates
```

pierwill: Or just "~a~ knowledge of".

src/logic/grammar.rkt
```diff
@@ -114,6 +114,16 @@
                (Quantifier KindedVarIds Goal)
                )
 
+  ;; A `Goal` (above) can be classified into the following categories.
```
pierwill:
```suggestion
  ;; A `Goal` can be classified into the following categories.
```

I actually think this is a bit distracting/confusing.

src/logic/hook.rkt
```diff
@@ -90,49 +86,58 @@
   )
 
 (define-metafunction formality-logic
-  ;; The "grammar" for predicates is just *any term* -- that's not very
-  ;; useful, and extension languages refine it. When matching against predicates,
-  ;; then, we can use this function to avoid matching on other kinds of goals.
-  is-predicate? : Env Goal -> boolean
+  ;; Categories the `Goal` according to the grammar defined by the upper layers
```
pierwill: "Upper" as in `formality-ty` and `formality-decl`? Is there a more descriptive way to phrase this that helps the reader understand the relationship between layers?

nikomatsakis: Hmm, I'm not sure! I feel like the "higher-level" layers are naturally going to be ones "closest" to the user's notation, but in this case, since this is the bottom layer, we could just write "other" layers too :)

nikomatsakis: I guess higher-level is somewhat more clear

PR #75, opened 2022-07-01
Use UserTy in impls and where clauses

voidc:
This changes the grammar for impls to `(impl KindedVarIds TraitId UserParameters for UserTy where WhereClauses ImplItems)`, making formality code more similar to Rust and thus improves readability.
It also changes the syntax of where clauses to use UserTy and resolves the ambiguity between `T: 'a` and `a': 'b` clauses by using the syntax `Lt_a >= Lt_b` for lifetime outlives clauses.
This has draft status as there are currently still some failing tests.

### DIFF
--- a/src/decl/decl-ok/crate-item.rkt
+++ b/src/decl/decl-ok/crate-item.rkt
@@ -5,6 +5,7 @@
          "../feature-gate.rkt"
          "../../logic/env.rkt"
          "../../logic/grammar.rkt"
+         "../../ty/user-ty.rkt"
          "trait-item.rkt"
          "impl-item.rkt"
          "wf-where-clause.rkt"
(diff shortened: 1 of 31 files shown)


nikomatsakis:
@voidc ah, cool! I was actually doing this too -- and I kind of came to the conclusion it might be the wrong approach, actually

nikomatsakis:
I have a branch that moves ALMOST ALL `Ty` to `UserTy`, but where-clauses were sticky, and it got me to thinking -- I think it might be worth investing in having "Rust-like declarations" that get lowered to the current Decls (maybe we even simplify the ones we have, though they're not bad).

nikomatsakis:
This would be a time to do some amount of elaboration, notably implied bounds.

voidc:
I agree that the Rust layer approach is more flexible and will close this PR.

PR #76, opened 2022-07-02
introduce Rust layer

nikomatsakis:
Experiment with adding a "Rust layer" that extends MIR. This gives a relatively natural syntax which gets "translated down" to the decl syntax. The idea is that we'll port our "integration tests" to this layer, and we'll also target it with @voidc's "rustc generation script". 

The other goal is to give a place to do "non-normative" expansions -- notably adding implied bounds.


nikomatsakis:
In retrospect, I think `(Vec < T >)` is inconsistent, and we should write it `(Vec[T])`.

PR #77, opened 2022-07-07
Expand MIR well-formedness rules and implement function input/output goals

voidc:
This makes the `type-check--return-ref-to-local` test pass.

### DIFF
--- a/src/mir/type-check-goal.rkt
+++ b/src/mir/type-check-goal.rkt
@@ -39,14 +39,15 @@
   #:mode (type-check-goal/inputs-and-outputs I O)
   #:contract (type-check-goal/inputs-and-outputs Γ GoalAtLocations)
 
-  [; FIXME: check that the "return local" has the same type as the one given in Γ
-   ;
-   ; FIXME: check that the types from the signature in Γ correspond to the types of the
-   ; corresponding local variables
-   ;
+  [(where/error (_ _ ((Ty_sigarg ..._n) -> Ty_sigret _ _) _) Γ)
+   (where/error ((_ Ty_locret _) (_ Ty_locarg _) ..._n _ ...) (local-decls-of-Γ Γ))
+   (where/error ((BasicBlockId_first _) _ ...) (basic-block-decls-of-Γ Γ))
+   (where/error Location (BasicBlockId_first @ 0))
    ; FIXME: check that the where-clauses in signature are well-formed?
    ----------------------------------------
-   (type-check-goal/inputs-and-outputs Γ ())
+   (type-check-goal/inputs-and-outputs Γ ((Location (Ty_sigret == Ty_locret))
+                                          (Location (Ty_sigarg == Ty_locarg)) ...
+                                          ))
    ]
   )
 
--- a/src/mir/type-check-goal.rkt
+++ b/src/mir/type-check-goal.rkt
@@ -39,9 +39,10 @@
   #:mode (type-check-goal/inputs-and-outputs I O)
   #:contract (type-check-goal/inputs-and-outputs Γ GoalAtLocations)
 
-  [(where/error (_ _ ((Ty_sigarg ..._n) -> Ty_sigret _ _) _) Γ)
+  [(where/error (_ _ ((Ty_sigarg ..._n) -> Ty_sigret _ _) _ _) Γ)
    (where/error ((_ Ty_locret _) (_ Ty_locarg _) ..._n _ ...) (local-decls-of-Γ Γ))
    (where/error ((BasicBlockId_first _) _ ...) (basic-block-decls-of-Γ Γ))
+   ; TODO: the goals should hold at "all locations"
    (where/error Location (BasicBlockId_first @ 0))
    ; FIXME: check that the where-clauses in signature are well-formed?
    ----------------------------------------
--- a/src/mir/type-of.rkt
+++ b/src/mir/type-of.rkt
@@ -2,6 +2,7 @@
 (require redex/reduction-semantics
          "grammar-extended.rkt"
          "../logic/substitution.rkt"
+         "../ty/user-ty.rkt"
          )
 (provide type-of/Place
          type-of/Rvalue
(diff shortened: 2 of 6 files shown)


src/mir/type-check-goal.rkt
```diff
@@ -39,14 +39,15 @@
...
-   ;
+  [(where/error (_ _ ((Ty_sigarg ..._n) -> Ty_sigret _ _) _) Γ)
+   (where/error ((_ Ty_locret _) (_ Ty_locarg _) ..._n _ ...) (local-decls-of-Γ Γ))
+   (where/error ((BasicBlockId_first _) _ ...) (basic-block-decls-of-Γ Γ))
+   (where/error Location (BasicBlockId_first @ 0))
    ; FIXME: check that the where-clauses in signature are well-formed?
    ----------------------------------------
-   (type-check-goal/inputs-and-outputs Γ ())
+   (type-check-goal/inputs-and-outputs Γ ((Location (Ty_sigret == Ty_locret))
+                                          (Location (Ty_sigarg == Ty_locarg)) ...
```
voidc, line 50: Should these be subtype goals?

src/mir/well-formed-mir.rkt
```diff
@@ -83,9 +119,35 @@
...
    ]
 
+  [(well-formed/Operand Γ Operand)
+   ----------------------------------------
+   (well-formed/Rvalue Γ (repeat Operand _))
+   ]
+
+  [(well-formed/Place Γ Place)
+   ----------------------------------------
+   (well-formed/Rvalue Γ (ref _ _ Place))
```
voidc: Should I check whether the lifetime is in the set of existential vars of the body?

src/mir/well-formed-mir.rkt
```diff
@@ -71,6 +86,27 @@
...
    ]
 
+  [(well-formed/Place Γ Place)
+   ----------------------------------------
+   (well-formed/Place Γ (* Place))
+   ]
+
+  [(well-formed/Place Γ Place)
+   ----------------------------------------
+   (well-formed/Place Γ (field Place _))
```
voidc, line 96: Should I check the existence of the `FieldId` here or is this better handled during type-checking?

nikomatsakis: Hmm. I think that's a type-check concern.

src/mir/well-formed-mir.rkt
nikomatsakis: We should check that the Lt is well-formed, which means either in the set of existential vars or one of the in-scope universal variables, I think.

src/mir/type-of.rkt
```diff
@@ -157,13 +158,13 @@
   [; ref
    (type-of/Place Γ Place Ty)
    ------------------------------------------
-   (type-of/Rvalue Γ (ref Lt () Place) (& Lt Ty))
+   (type-of/Rvalue Γ (ref Lt () Place) (rigid-ty (ref ()) (Lt Ty)))
```
nikomatsakis, line 161: `   (type-of/Rvalue Γ (ref Lt () Place) (user-ty (& Lt Ty)))`, maybe?

src/mir/type-of.rkt
```diff
@@ -157,13 +158,13 @@
...
    ]
 
   [; ref-mut
    (type-of/Place Γ Place Ty)
    ------------------------------------------
-   (type-of/Rvalue Γ (ref Lt mut Place) (&mut Lt Ty))
+   (type-of/Rvalue Γ (ref Lt mut Place) (rigid-ty (ref mut) (Lt Ty)))
```
nikomatsakis, line 167: as above

src/mir/type-check-goal.rkt
nikomatsakis: Equality is better, I think, though subtyping probably suffices.

nikomatsakis: I am reminded that, in rustc, we also have a concept of "all locations" we use sometimes, and I think maybe this would be an instance where it is appropriate. But that's probably a TODO to be filed later.

src/rust/test/type-check--return-ref-to-local.rkt
```diff
@@ -9,27 +9,21 @@
...
-                  [(fn foo[(lifetime a)] () -> (& a u32)
+                  [(fn foo[(lifetime a)] () -> (& a i32)
                        where []
                        { ∃ [(lifetime ?a)
                             (lifetime ?b)
                             ]
-                           ([(_0 (user-ty (& ?a u32)) mut)
-                             (_1 (user-ty u32) ()) ; the variable `x`
+                           ([(_0 (user-ty (& ?a i32)) mut)
+                             (_1 (user-ty i32) ()) ; the variable `x`
```
nikomatsakis, line 26: is it because `const` has type `i32`?

src/mir/type-of.rkt
voidc: This won't work since we currently can't turn a `Ty` into a `UserTy`. Originally, there existed a `&` metafunction that created reference types.

src/mir/type-check-goal.rkt
voidc: Okay, I added a TODO.

src/rust/test/type-check--return-ref-to-local.rkt
voidc: Yes

src/mir/well-formed-mir.rkt
voidc: Done. Do we also need to check well-formedness of the types of `LocalDecls`?

nikomatsakis: Yes, we ought to prove that, but let's file an issue for it, because i'm working a lot on our approach to well-formedness right now and some of the helper functions I think you need don't exist yet =)

src/mir/type-of.rkt
nikomatsakis: Oh, I see, yes. Hmm =) Maybe we want a `ref-ty` helper or something, but this is fine.

nikomatsakis (approved):

src/mir/well-formed-mir.rkt
nikomatsakis: Created https://github.com/nikomatsakis/a-mir-formality/issues/78

PR #81, opened 2022-07-13
split mir into body, check layers

nikomatsakis:
Body: defines the MIR and the type-check / borrow-check

Check: defines the overall safety check


nikomatsakis:
(cc @voidc -- not sure if you've got outstanding edits)

PR #87, opened 2022-08-17
Add Rust to mir-formality converter

voidc:
This PR adds the code of the [mir-formality-gen](https://github.com/voidc/mir-formality-gen) converter.
The tool interfaces with rustc to translate the internal `rustc_middle` representation to mir-formality syntax.
Many of Rust's features can not yet be translated as they are not yet modeled in mir-formality.
A pair of Rust code and converted mir-formality code is included in `gen/tests/input`.


voidc:
For the included example, the tool prints the following list of Rust constructs it fails to translate:
```
unknown crate item: #0
unknown crate item: std#2
unknown rvalue: ()
unknown rvalue: foo::<T> as for<'a, 'b> fn(&'a &'b (), &'b T) -> &'a T (Pointer(ReifyFnPointer))
unknown stmt: FakeRead(ForLet(None), _2)
unknown stmt: AscribeUserType(_2, o, UserTypeProjection { base: UserType(1), projs: [] })
unknown const: const {alloc1: &&&()}
unknown stmt: FakeRead(ForLet(None), _2)
unknown const: bad::<i32>
unknown stmt: FakeRead(ForLet(None), _1)
unknown const: const ()
```

gen/src/gen.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    pub fn emit_ident(&self, ident: &Symbol) -> String {
+        // Racket special characters: ( ) [ ] { } " , ' ` ; # | \
+        ident.as_str().replace('\'', Self::LIFETIME_MARKER)
+    }
+
+    pub fn emit_def_path(&self, def_id: DefId) -> String {
+        let def_path = self.tcx.def_path_str(def_id);
+        match def_path.as_str() {
```
nikomatsakis, line 53: Ideally we would ask the compiler to inspect the `#[lang]` annotation, but this will do for now...

nikomatsakis (approved):
This looks great!

PR #95, opened 2022-09-06
Expand body grammar

voidc:
This PR expands the body grammar, adding aggregate and cast rvalue variants as well as supporting more constant values. Besides the racket definitions, the formality generator was adapted as well. As a result, the issue25860 example can now be fully translated. The well-formedness and type checking of the new constructs is not yet fully implemented. Also, I want to add some more tests.

### DIFF
--- a/racket-src/body/type-of.rkt
+++ b/racket-src/body/type-of.rkt
@@ -46,7 +46,7 @@
    ;s
    ; extract the name of the singular variant
    (place-type-of Γ Place (rigid-ty AdtId Parameters) ())
-   (where (struct AdtId _ where _ ((VariantId _))) (decl-of-adt Γ AdtId))
+   (where (struct AdtId _ where _ ((VariantId _))) (find-adt Γ AdtId))
    (field-tys Γ AdtId Parameters VariantId (_ ... (FieldId Ty_field) _ ...))
    ----------------------------------
    (place-type-of Γ (field Place FieldId) Ty_field ())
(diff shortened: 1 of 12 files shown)


racket-src/body/type-of.rkt
```diff
@@ -91,19 +92,61 @@
...
+   ------------------------------------------
+   (type-of/Constant Γ true (user-ty bool))]
+
+  [; false
+   ------------------------------------------
+   (type-of/Constant Γ false (user-ty bool))]
+
+  [; function
+   (where/error (fn _ KindedVarIds (Ty_arg ...) -> Ty_ret _ _ _) (find-fn Γ FnId))
+   (where/error (KindedVarId_subst ..._n KindedVarId_other ...) KindedVarIds)
```
nikomatsakis, line 129: Wait, what is this `_n` exactly? I don't understand why we want to take the first `N` generic parameters.

nikomatsakis (commented):
Gave a quick skim -- looks good, modulo one issue.

racket-src/body/type-of.rkt
voidc: I assume that the first `n` parameters of the function are early-bound and the rest are late-bound -- not sure if this is the right behavior. Thus a function pointer `(fn-ptr foo [T])` to `fn foo[(type T) (lifetime %a) (lifetime %b)] ((& %a (& %b ())) (& %b T)) -> (& %a T)` would get a type like `(∀[(lifetime %a) (lifetime %b)] (rigid-ty ...))`.

PR #97, opened 2022-09-14
add a fn-def type and adjust type of fn constants

nikomatsakis:
but we're not really testing it in any useful way.

cc @voidc

### DIFF
--- a/racket-src/body/type-of.rkt
+++ b/racket-src/body/type-of.rkt
@@ -125,14 +125,35 @@
    (type-of/Constant Γ false (user-ty bool))]
 
   [; function
-   (where/error (fn _ KindedVarIds (Ty_arg ...) -> Ty_ret _ _ _) (find-fn Γ FnId))
-   (where/error (KindedVarId_subst ..._n KindedVarId_other ...) KindedVarIds)
-   (where/error Substitution (create-substitution (KindedVarId_subst ...) (Parameter ...)))
-   (where/error (Ty_argsubst ...) ((apply-substitution Substitution Ty_arg) ...))
-   (where/error Ty_retsubst (apply-substitution Substitution Ty_ret))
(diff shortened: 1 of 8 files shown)


nikomatsakis:
Huh, tests were passing locally for me, I thought.

racket-src/body/type-of.rkt
```diff
@@ -125,14 +125,35 @@
...
+
+   ; Split the generics from the function into early/late
+   (where/error (KindedVarId_early ..._n KindedVarId_late ...) KindedVarIds)
+
+   ; Create a substitution from early-bound parameters to the values from user
+   (where/error Substitution_early (create-substitution (KindedVarId_early ...) (Parameter ...)))
+   (where/error Biformulas_early (apply-substitution Substitution_early Biformulas))
+
+   ; Apply that substitution
+   (where/error Ty_fn (rigid-ty (fn-def FnId) (Parameter ... KindedVarId_late ...)))
```
voidc: I think we need to extract the `VarId` from the `KindedVarId_late` here

voidc: I fixed this here: https://github.com/nikomatsakis/a-mir-formality/compare/ty-fn-def...voidc:a-mir-formality:ty-fn-def

nikomatsakis: oh, yeah, that's correct

nikomatsakis:
I don't understand why CI is failing :(

nikomatsakis:
Ah, but it actually *started* failing when I merged #95, it appears. Curious.

nikomatsakis:
The PR itself seemed fine, but browsing the actions, I see this: https://github.com/nikomatsakis/a-mir-formality/actions/runs/3054378958/jobs/4926199312

nikomatsakis:
It works locally, I'm going to merge, but we have to figure this out at some point!

PR #102, opened 2022-09-17
Solutions not schemes

nikomatsakis:
This PR is working towards creating a recursive solver and a query-based model for mir-formality. It introduces the query infrastructure and ports a number of tests to use it. The core solver interface is now in the `logic:solve-query` metafunction:

```scheme
  ;; Compute the solutions to the given query and return them.
  logic:solve-query : Hook QueryGoal -> Solutions
```

A `QueryGoal` encapsulates all the existential/universal variables, hypotheses, as well as the goal we are trying to prove:

```scheme
  ;; `QueryGoal` -- A *query term* is a goal packaged up with variables that
  ;; are referenced in the goal. Each variable carries its quantifier (∀ vs ∃)
  ;; along with a universe.
  ;;
  ;; Given a query, one can create an environment that has all the given variables
  ;; and universes in scope. This can be used to solve the canonical term and produce a result.
  ;;
  ;; The result of a query is a `Solution` (or set of `Solutions`).
  (QueryGoal ::= (?- VarBinders (implies Hypotheses Goal)))
```

and the solution gives back the conditions in which the query-goal is provable:

```scheme
  ;; A *solution* consists of inference variables and relations between them.
  ;; When the solver decides something is provable, it does so modulo the "solution",
  ;; which may contain various relations that were not proven, and which must be
  ;; proven at another level. Note that these are 'first-order' relations, though.
  (Solutions ::= (Solution ...))
  (Solution ::= (:- ∃VarBinders (Substitution Relations)))
```


nikomatsakis:
It works locally, I'm going to merge.

PR #104, opened 2022-09-21
Type-check function calls and reification casts

### LINKED ISSUE #100: Type-checking calls: when there is a CALL in the MIR, we need to "unpack" the fn type to get to something callable. e.g. if the type of the thing-to-be-called is `forall<'a> (T => F)`, then we should instantiate `'a` with an inference variable, add `T` as an obligation, and finally get the signature from `F`.

### LINKED ISSUE #101: Type-checking reification casts
Similar to #100, when we reify a fn-def into a fn-ptr, we need to do a vaguely similar process. 

One option would be to take a `for<....> (P => F)` type (where `F` is a fn-def) and "convert" it to a fn-ptr type, basically by replacing the `F` with a `(fn-ptr (Ty_arg ...) -> Ty_ret)` (with suitable ABI and the like).

and then do subtyping between that and the target of the cast.

This is very similar to the original code that @voidc that I replaced in #97.

voidc:
This implements a function to get the signature from a function type and uses it to type-check function calls and reification casts in MIR. 

closes #100
closes #101

### DIFF
--- a/racket-src/body/type-check-goal.rkt
+++ b/racket-src/body/type-check-goal.rkt
@@ -140,10 +140,11 @@
    (fn-ty-signature Γ Ty_fn (∀ KindedVarIds (implies Biformulas ((Ty_arg ...) -> Ty_ret))))
    (where/error number_args ,(length (term (Ty_arg ...))))
    (where/error Ty_fnptr (rigid-ty (fn-ptr "Rust" number_args) (Ty_arg ... Ty_ret)))
+   (where/error Ty_reified (∀ KindedVarIds (implies Biformulas Ty_fnptr)))
    ----------------------------------------
    (type-check-goal/Rvalue Γ
                            (cast Operand as Ty_target)
-                           (∀ KindedVarIds (implies Biformulas (Ty_fnptr <= Ty_target))))
+                           (Ty_reified <= Ty_target))
    ]
 
   [----------------------------------------
@@ -157,18 +158,18 @@
   #:mode (fn-ty-signature I I O)
   #:contract (fn-ty-signature Γ Ty Signature)
 
-  [(fn-ty-signature Γ Ty (∀ () (implies Biformulas (Tys_args -> Ty_ret))))
+  [(fn-ty-signature Γ Ty (∀ [KindedVarId_2 ...] (implies Biformulas (Tys_args -> Ty_ret))))
    ----------------------------------------
    (fn-ty-signature Γ
-                    (∀ KindedVarIds Ty)
-                    (∀ KindedVarIds (implies Biformulas (Tys_args -> Ty_ret))))
+                    (∀ [KindedVarId_1 ...] Ty)
+                    (∀ [KindedVarId_1 ... KindedVarId_2 ...] (implies Biformulas (Tys_args -> Ty_ret))))
    ]
 
-  [(fn-ty-signature Γ Ty (∀ () (implies () (Tys_args -> Ty_ret))))
+  [(fn-ty-signature Γ Ty (∀ () (implies [Biformula_2 ...] (Tys_args -> Ty_ret))))
    ----------------------------------------
    (fn-ty-signature Γ
-                    (implies Biformulas Ty)
-                    (∀ () (implies Biformulas (Tys_args -> Ty_ret))))
+                    (implies [Biformula_1 ...] Ty)
+                    (∀ () (implies [Biformula_1 ... Biformula_2 ...] (Tys_args -> Ty_ret))))
    ]
 
   [----------------------------------------
--- a/racket-src/body/type-check-goal.rkt
+++ b/racket-src/body/type-check-goal.rkt
@@ -157,10 +157,24 @@
   #:mode (fn-ty-signature I I O)
   #:contract (fn-ty-signature Γ Ty Signature)
 
+  [(fn-ty-signature Γ Ty (∀ () (implies Biformulas (Tys_args -> Ty_ret))))
+   ----------------------------------------
+   (fn-ty-signature Γ
+                    (∀ KindedVarIds Ty)
+                    (∀ KindedVarIds (implies Biformulas (Tys_args -> Ty_ret))))
+   ]
+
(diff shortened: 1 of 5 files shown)


racket-src/body/type-check-goal.rkt
```diff
@@ -135,9 +135,43 @@
...
 
+  [; type-check ReifyFnPointer casts
+   (type-of/Operand Γ Operand Ty_fn)
+   (fn-ty-signature Γ Ty_fn (∀ KindedVarIds (implies Biformulas ((Ty_arg ...) -> Ty_ret))))
+   (where/error number_args ,(length (term (Ty_arg ...))))
+   (where/error Ty_fnptr (rigid-ty (fn-ptr "Rust" number_args) (Ty_arg ... Ty_ret)))
+   (where/error Ty_reified (∀ KindedVarIds (implies Biformulas Ty_fnptr)))
+   ----------------------------------------
+   (type-check-goal/Rvalue Γ
+                           (cast Operand as Ty_target)
```
nikomatsakis, line 146: OK, so, `Ty_target` should *always* be a fn pointer type, right? I think I expect this to look a bit more like this

```
[(is-fn-pointer Ty_target) ; some kind of simple check
 (type-of/Operand Γ Operand Ty_fn)
 (fn-cast-goal Γ Ty_fn Ty_target Goal)
 ----------------------------------------
 (type-check-goal/Rvalue Γ (cast Operand as Ty_target) Goal)
 ]
```

where `fn-cast-goal` has a recursion shape like this

```
[(fn-cast-goal Γ Ty Ty_target)
 ----------------------------------------
 (fn-cast-goal Γ (∀ KindedVarIds Ty) Ty_target (∀ KindedVarIds Goal))

[(fn-cast-goal Γ Ty Ty_target)
 ----------------------------------------
 (fn-cast-goal Γ (implies Biformulas Ty) Ty_target (implies Biformulas Goal))

[(where/error (fn FnId KindedVarIds_fn (Ty_arg ...) -> Ty_ret _ _ _) (find-fn Γ FnId))
 (where/error Substitution (create-substitution KindedVarIds_fn Parameters))
 (where/error (Ty_argsubst ...) ((apply-substitution Substitution Ty_arg) ...))
 (where/error Ty_retsubst (apply-substitution Substitution Ty_ret))
 (where/error Ty_fnptr (rigid-ty (fn-ptr "Rust" number_args) [Ty_argsubst ... Ty_retsubst]))
 ----------------------------------------
 (fn-cast-goal Γ (rigid-ty (fn-def FnId) Parameters) Ty_target (Ty_fnptr <= Ty_target))
```

This way we aren't hardcoding the type structure so much, we just translate it to goals as we pass through it.

nikomatsakis (changes requested):
WDYT?

racket-src/body/type-check-goal.rkt
voidc: I moved the binders of the operand's type `Ty_fn` to the goal. Considering that `Ty_target` can have its own `∀` binders, does this give the correct behavior? Regarding not hardcoding the type structure, I made some changes to `fn-ty-signature`. Does that already accomplish what you had in mind?

nikomatsakis: > Considering that Ty_target can have its own ∀ binders, does this give the correct behavior? 

Hmm, I thought so, but now I'm realizing maybe this wasn't quite right. What we want may actually be more like

```
(reified-fn-type Gamma Ty_in Ty_out)
```

...which does the translation, still I think in the recursive way I described, but changing the innermost `fndef` to `fnptr`, and then...

```
(reified-fn-type Gamma Ty_source Ty_reified)
(where/error Goal (Ty_reified <= Ty_target))
```

nikomatsakis: In particular, the structure I described proves that the cast works for *every possible value* of the input lifetimes, but actually we only need it to hold for *some value*. The code as I wrote it should error out, for example, if you do this cast

```rust
fn foo(x: &u8) {}

let x: fn(&'static u8) = foo;
```

(you could try adding a test)

but the version above would work fine (since it would reify `foo` to `fn(&u8)`, and then this is a subtype of `fn(&'static u8)`).

racket-src/body/type-check-goal.rkt
```diff
@@ -135,9 +135,56 @@
...
    ]
 
   )
 
+(define-judgment-form
+  formality-body
+  #:mode (fn-ty-signature I I O)
+  #:contract (fn-ty-signature Γ Ty Signature)
+
+  [(fn-ty-signature Γ Ty (∀ () (implies Biformulas (Tys_args -> Ty_ret))))
```
nikomatsakis: Ideally, we would not hard-code that this recursive judgment is going to yield back a `()` list of ids, but instead "merge" these signatures, like so

```
[(fn-ty-signature Γ Ty (∀ [KindedVarId_1 ...] (implies Biformulas_1 (Tys_args -> Ty_ret))))
 ------
 (fn-ty-signature Γ
                    (∀ [KindedVarId_0 ...] Ty)
                    (∀ [KindedVarId_0 ... KindedVarId_1 ...] (implies Biformulas_1 (Tys_args -> Ty_ret))))
 ]
```

this would allow us to have a type like `(ForAll [...] (ForAll [...] ...))` and still construct a sensible signature for it.

racket-src/body/type-check-goal.rkt
```diff
@@ -135,9 +135,56 @@
...
+   ----------------------------------------
+   (fn-ty-signature Γ
+                    (∀ KindedVarIds Ty)
+                    (∀ KindedVarIds (implies Biformulas (Tys_args -> Ty_ret))))
+   ]
+
+  [(fn-ty-signature Γ Ty (∀ () (implies () (Tys_args -> Ty_ret))))
```
nikomatsakis: Same here, we would ideally get back a `(ForAll KindedVarIds (implies Biformulas_1 (...))` and then concatenate that with the biformulas we "peeled off"

nikomatsakis (changes requested):
Looking good! One last nit.

nikomatsakis (approved):

nikomatsakis:
Woohoo!

Issue #99, opened 2022-09-14
Well-formedness rules: We need tests that fn-def types are well-formed (and maybe code?). To be well-formed, the where-clauses on the fn must be met. (That said, in the MIR type checker, I am currently adding in implications that will basically always make this true.)


nikomatsakis:

shua:
I may have addressed this with #181

PR #106, opened 2022-09-24
Add well-formed-fn predicate

voidc:
Addresses #99. This only adds a way to check that `fn-def` types are well-formed but does check the well-formedness of e.g. function constants.

### DIFF
--- a/racket-src/decl/decl-to-clause/trait-item.rkt
+++ b/racket-src/decl/decl-to-clause/trait-item.rkt
@@ -14,9 +14,82 @@
   ;; * The invariants that hold
   trait-item-decl-rules : CrateDecls CrateId (TraitId KindedVarIds_trait) TraitItem -> (Clauses Invariants)
 
-  [;; For a method declared in a trait: currently no logical rules are created.
-   (trait-item-decl-rules CrateDecls CrateId (TraitId KindedVarIds_trait) FnDecl)
-   ([] [])
+  [;; For a method declared in the trait Bar with parameter Self, like the following:
+   ;;
+   ;;     fn foo<T>(x: T) -> T where T: Ord { ... }
+   ;;
+   ;; We generate the following clauses
+   ;;
+   ;;     (∀ [(type Self) (type T)]
+   ;;         (well-formed-fn foo[Self T]) :-
+   ;;            (well-formed (type Self))
+   ;;            (well-formed (type T))
+   ;;            (is-implemented (Bar Self)))
+   ;;            (is-implemented (Ord T)))
+   ;;
+   ;; And the following invariants local to the crate C:
+   ;;
+   ;;     (∀ [(type Self) (type T)]
+   ;;         (well-formed (type foo[Self T])) => (is-implemented (Ord T)))
+   ;;
+   ;; And the following global invariants:
+   ;;
+   ;;     (∀ [(type Self) (type T)]
+   ;;         (well-formed (type foo[Self T])) => (&& (well-formed (type Self))
+   ;;                                                 (well-formed (type T))))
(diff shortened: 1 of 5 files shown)


racket-src/decl/decl-to-clause/trait-item.rkt
```diff
@@ -14,9 +14,82 @@
...
+   ;;
+   ;;     fn foo<T>(x: T) -> T where T: Ord { ... }
+   ;;
+   ;; We generate the following clauses
+   ;;
+   ;;     (∀ [(type Self) (type T)]
+   ;;         (well-formed-fn foo[Self T]) :-
+   ;;            (well-formed (type Self))
+   ;;            (well-formed (type T))
+   ;;            (is-implemented (Bar Self)))
```
voidc, line 28: Do we need a `(well-formed (type foo[Self T])) => (is-implemented (Bar Self)))` invariant?

racket-src/decl/decl-to-clause/trait-item.rkt
```diff
@@ -14,9 +14,82 @@
...
+  [;; For a method declared in the trait Bar with parameter Self, like the following:
+   ;;
+   ;;     fn foo<T>(x: T) -> T where T: Ord { ... }
+   ;;
+   ;; We generate the following clauses
+   ;;
+   ;;     (∀ [(type Self) (type T)]
+   ;;         (well-formed-fn foo[Self T]) :-
+   ;;            (well-formed (type Self))
+   ;;            (well-formed (type T))
```
voidc, line 27: I saw that associated types do not require well-formedness of their parameters. Why is that? Is it correct to require it for associated functions?

racket-src/decl/well-formed/parameter.rkt
```diff
@@ -125,6 +125,10 @@
    [(well-formed (Type Ty)) ...]
    ]
 
+  [(well-formed-subgoals-for-ty CrateDecls (rigid-ty (fn-def FnId) Parameters))
```
nikomatsakis, line 128: Hmm. This is orthogonal from this PR, but if this function is a method inside a trait, we probably shouldn't just be using a flat `FnId`, right? That doesn't let us distinguish two traits with the same functions.

We probably want to have `(fn-def FnName)` where

```
(FnName ::= FnId (TraitId FnId))
```

much like `AliasName`.

racket-src/decl/decl-to-clause/trait-item.rkt
nikomatsakis: I think that is an oversight. We should treat the item generics in the same way as the impl generics.

racket-src/decl/decl-to-clause/trait-item.rkt
nikomatsakis: We don't need this today, but I could see it becoming relevant in the future. Right now user's can't name these types explicitly, so they don't show up -- ah, well, hmm, maybe they can show up via inference! Something like this:

```rust
fn foo<T>() { }

fn bar() {
    let x = Some(foo).map(|f| ...);
}
```

inside that closure, `f` has a `FnDef` type. So in principle `(well-formed f)` would appear in its closure environment, and we could infer things based on that. Of course, anything we could infer should also be provable in the environment we inherit from `bar`. 

I am inclined to say NO invariants to start, but we should open an issue and compare against rustc I think.

nikomatsakis (commented):
Nice. I left a few answers, but also an orthogonal concern (we ought to file an issue for that) concerning how we name fn-def types.

nikomatsakis:
I'm going to close this because we're moving to the Rust version

PR #108, opened 2022-09-26
Split CI racket tests according to directories.

crlf0710:
This reduces CI run duration from `~17m 43s` to `~8m 8s` with more parallelisms. The downside is that if a new module is added beside `logic, ty, decl, body, check, rust` in the future, ci configuration needs to be updated for its tests to run.


nikomatsakis:
Seems worth it! We'll have to remember that, though.

PR #118, opened 2022-11-28
Rewrite It In Rust

nikomatsakis:
This is a long-running effort to rewrite in Rust. It's not at any particular stopping point but it seems like time to merge it down.


Cargo.toml
```diff
@@ -13,3 +13,16 @@ pretty_assertions = "1.3.0"
...
+    "crates/formality-macros",
+    "crates/formality-core",
+    "crates/formality-types",
+    "crates/formality-logic",
+    "crates/formality-decl",
+    "crates/formality-mir",
+    "crates/formality-check",
+    "crates/formality-rust",
+    "crates/formality-main",
+]
```
bjorn3: nit: missing trailing newline

nikomatsakis: fixed :)

nikomatsakis:
just gonna merge this :)

PR #127, opened 2023-06-28
Add const generic params

oli-obk:
Best reviewed commit by commit.

Before I continue down this road, does this work look in the spirit of how you're imagining a-mir-formality to go?

### DIFF
--- a/crates/formality-types/src/grammar/ty.rs
+++ b/crates/formality-types/src/grammar/ty.rs
@@ -795,12 +795,12 @@ impl VarSubstitution {
         t.substitute(&mut |v| Some(self.map.get(&v)?.upcast()))
     }
 
-    pub fn map_var(&self, v: Variable) -> Option<Variable> {
-        self.map.get(&v).copied()
+    pub fn map_var(&self, v: &Variable) -> Option<Variable> {
+        self.map.get(v).copied()
     }
 
-    pub fn maps_var(&self, v: Variable) -> bool {
-        self.map.contains_key(&v)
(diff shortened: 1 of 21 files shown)


crates/formality-types/src/grammar/ty.rs
```diff
@@ -113,7 +117,7 @@ impl UpcastFrom<Ty> for TyData {
     }
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
nikomatsakis, line 123: Why make Variable `!Copy`?

nikomatsakis (approved):
This looks generally right but I'm not sure why you are making `Variable: !Copy`.

nikomatsakis:
In particular, I have some thoughts about how to handle specifying the type of constants. I think it wants to be stored in the environment as a predicate.

oli-obk:
Yea, I have recently (Friday) looked into removing the type of constants from them, and if we could first show that this works in a-mir-formality, that would be great. I'll investigate if I can do this PR with the predicate instead, but that has its own problems (that are easier to investigate in a-mir-formality than in rustc)

oli-obk:
> In particular, I have some thoughts about how to handle specifying the type of constants. I think it wants to be stored in the environment as a predicate. (I am guessing this may be related to my question above.)

I did this in https://github.com/rust-lang/a-mir-formality/pull/130

We still need to store a type for rigid constants, but we don't need the type in ParameterKind anymore.

PR #128, opened 2023-06-29
Implement unsafe trait support

### LINKED ISSUE #50: Add support for unsafe traits, impls
We need to support *unsafe* traits and impls. The `rust:Send` and `rust:Sync` traits should be defined as unsafe.

yoshuawuyts:
Closes #50. This implements support for `unsafe trait` and `unsafe impl`. However, I'm not yet super comfortable with the parser - so I've implemented it using an enum which means that non-unsafe trait decls and impls are now written: `safe trait` and `safe impl` respectively.

Is the right approach to instead manually implement the `Parser` trait? I have a feeling that the MIR syntax is intended to be unambiguous, so is that even something we want to do? Or should every trait decl and impl be explicit about whether it's unsafe or not? Any guidance here would be appreciated!

~~_Best reviewed one commit at a time_~~

### DIFF
--- a/crates/formality-prove/src/decls.rs
+++ b/crates/formality-prove/src/decls.rs
@@ -125,10 +125,21 @@ pub struct NegImplDeclBoundData {
 }
 
 /// Mark a trait or trait impl as `unsafe`.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Safety {
     Safe,
     Unsafe,
+}
+
+// NOTE(yosh): `Debug` is currently used to print error messages with. In order
+// to not print `safe impl` / `safe trait` where none is written, we leave the impl blank.
+impl std::fmt::Debug for Safety {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Safe => write!(f, ""),
+            Self::Unsafe => write!(f, "unsafe"),
+        }
+    }
 }
 
 impl Term for Safety {}
--- a/crates/formality-prove/src/decls.rs
+++ b/crates/formality-prove/src/decls.rs
@@ -95,8 +95,9 @@ impl Decls {
     }
 }
 
-#[term(impl $binder)]
+#[term($safety impl $binder)]
 pub struct ImplDecl {
+    pub safety: Safety,
     pub binder: Binder<ImplDeclBoundData>,
 }
 
@@ -106,8 +107,9 @@ pub struct ImplDeclBoundData {
     pub where_clause: Wcs,
 }
 
-#[term(impl $binder)]
+#[term($safety impl $binder)]
 pub struct NegImplDecl {
+    pub safety: Safety,
     pub binder: Binder<NegImplDeclBoundData>,
 }
 
@@ -117,8 +119,15 @@ pub struct NegImplDeclBoundData {
     pub where_clause: Wcs,
 }
 
-#[term(trait $id $binder)]
+#[term]
+pub enum Safety {
+    Safe,
+    Unsafe,
+}
+
+#[term($safety trait $id $binder)]
 pub struct TraitDecl {
+    pub safety: Safety,
     pub id: TraitId,
     pub binder: Binder<TraitDeclBoundData>,
 }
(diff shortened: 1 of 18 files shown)


yoshuawuyts:
I've instead hand-rolled the parser, removing the need for `safe trait` and `safe impl` notation - only keeping `unsafe trait` and `unsafe impl` where needed. I believe this should make it more natural  to write.

crates/formality-prove/src/decls.rs
```diff
@@ -117,8 +119,15 @@ pub struct NegImplDeclBoundData {
     pub where_clause: Wcs,
 }
 
-#[term(trait $id $binder)]
+#[term]
```
nikomatsakis: Hmm, so if we wanted to do 

```rust
trait Foo { }
```

vs

```rust
unsafe trait foo { }
```

we have two ways to do it:

* Implement the parsing traits manually for `Safety` 
* Or extend the procedural macro to accommodate this case

I'm inclined to do the latter, as I think it will come up. Something like setting a "default" variant.

crates/formality-prove/src/decls.rs
```diff
@@ -125,12 +125,23 @@ pub struct NegImplDeclBoundData {
 }
 
 /// Mark a trait or trait impl as `unsafe`.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Safety {
     Safe,
     Unsafe,
 }
```
nikomatsakis: I think you could get this same effect with 

```rust
#[term]
pub enum Safety {
    #[grammar()]
    Safe,
    #[grammar(unsafe)]
    Unsafe,
}
```

That said, I didn't recommend it earlier because I thought we probably wanted to keep `safe` in the debug output. I'm torn between:

* Accept (and print) both `safe` and `unsafe` but have a default when parsing -- what I was originally thinking
* Never print safe -- wht you are doing

The former seems more explicit and clear, but it departs a bit from Rust surface syntax. I still think that's the way to go though, we do similar things in a few other places. I do think it's important to not require users to write `safe trait` in tests though.

nikomatsakis (changes requested):

nikomatsakis:
What do you think about [this question](https://github.com/rust-lang/a-mir-formality/pull/128#discussion_r1257435082), @yoshuawuyts ?

crates/formality-prove/src/decls.rs
yoshuawuyts: > I'm inclined to do the latter, as I think it will come up. Something like setting a "default" variant.

I think that's reasonable; I honestly don't love the custom parsing strategy I took in this PR, and I'm happy to instead change the proc macro.

crates/formality-prove/src/decls.rs
yoshuawuyts: Yeah, I agree that that seems like the way to go. I initially had `safe trait Foo`-notation everywhere, and that felt really clumsy. But just having it as part of debug output seems like it would be good!

tests/coherence_orphan.rs
```diff
@@ -31,7 +31,7 @@ fn test_orphan_neg_CoreTrait_for_CoreStruct_in_Foo() {
     expect_test::expect![[r#"
         Err(
             Error {
-                context: "orphan_check_neg(impl <> ! CoreTrait < > for (rigid (adt CoreStruct)) where [] {})",
+                context: "orphan_check_neg( impl <> ! CoreTrait < > for (rigid (adt CoreStruct)) where [] {})",
```
nikomatsakis, line 34: hmm this is annoying :)

nikomatsakis (changes requested):
Hmm, this is exposing a shortcoming of the pretty printer, but it's not obvious to me how to fix it. Ideally we'd skip the separator if the field before printed nothing, but it's kind of tricky to know when that might be true. Well, we could generate a `String` and then check its length...? Anyway, I guess that's a distinct PR. =) The code to modify would be in [debug.rs](https://github.com/rust-lang/a-mir-formality/blob/main/crates/formality-macros/src/debug.rs). I looked though and didn't see a *tiny* diff that would work.

nikomatsakis:
I think I started to lay the groundwork to handle this nicely now. We have `$?x` patterns. We could implement `Default` to safe/unsafe and then we could modify the debug impls to check if the value is the default value and skip over it.

yoshuawuyts:
I think I should probably throw in the towel on this patch for now; I currently don't have the bandwidth to see this through. If someone wants to take this over please feel free to run with this. Happy to close this if that's preferred too.

nikomatsakis:
@lqd said they were up for it

nikomatsakis:
Closing in favor of #155

PR #131, opened 2023-07-26
Move some tests to ui_test

oli-obk:
I haven't ported all tests yet because

* [ ] some tests like `test_overlap_alias_not_normalizable` are basically revision tests, but we have no way to handle --cfg flags passed to the formality binary yet.
* [ ] some tests like `test_mirror_normalizes_u32_to_u32` don't just check if a program compiles, but whether a where clause holds/what its outputs are

It works great, and with the control we have over command line flags, we probably resolve all of these issues.

The file extension is a bit annoying because you need to copy paste it and can't type it, but 🤷 it's very cute.

Also featuring: ui_test with progress bars (you can't see it because the tests finish so fast).


lqd:
does `.🔬` work on all OSes ? :)

oli-obk:
It works on windows Mac and linux. That's not all, but covers the host distros we can test

nikomatsakis (approved):

nikomatsakis:
Not sure about the microscope file extension, seems like it might be annoying after a while. I wonder about `.fy`, `.fty`, or even `.formality` -- or `.👔` ??

nikomatsakis:
Gonna merge for now.

PR #137, opened 2023-09-10
Add ui tests for functions

obeis:
Updates #99

cc @oli-obk

### DIFF
--- a/tests/ui/fn/ok.🔬
+++ b/tests/ui/fn/ok.🔬
@@ -1,0 +1,20 @@
+// Test functions, function's arguments, and function's returns
+//@check-pass
+[
+    crate Foo {
+        // fn simple_fn() {}
+        fn simple_fn<>() -> () where [] { trusted }
+
+        // fn one_arg<T>(_: T) {}
+        fn one_arg<ty T>(T) -> () where [] { trusted }
+
+        // fn one_ret<T>(_: T) {}
+        fn one_ret<ty T>() -> T where [] { trusted }
+
+        // fn arg_ret<T, U>(_: T) -> U {}
+        fn arg_ret<ty T, ty U>(T) -> U where [] { trusted }
+
+        // fn multi_arg_ret<T, Y, U, I>(_: T, _: Y) -> (U, I) {}
+        fn multi_arg_ret<ty T, ty Y, ty U, ty I>(T, Y) -> (U, I) where [] { trusted }
+    }
+]
--- a/crates/formality-check/src/fns.rs
+++ b/crates/formality-check/src/fns.rs
@@ -3,9 +3,7 @@ use formality_rust::{
     grammar::{Fn, FnBoundData},
     prove::ToWcs,
 };
-use formality_types::{
-    grammar::{Fallible, Wcs},
-};
+use formality_types::grammar::{Fallible, Wcs};
 
 use crate::Check;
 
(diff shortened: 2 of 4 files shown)


tests/ui/fn/ok.stderr
```diff
@@ -0,0 +1,29 @@
...
+       
+               fn is_true(v: bool) -> bool {}
+       
+               fn add(x: i64, y: i64) -> i64 {}
+       
+               fn return_multi(a: i32, b: i32) -> (i32, i32) {}
+           }
+       ]
+       
+    1: failed to parse [
```
compiler-errors: This is a parser error. Have you looked at the other tests that demonstrate the difference in a-mir-formality syntax?

tests/ui/fn/ok.🔬
```diff
@@ -0,0 +1,13 @@
+[
+    crate Foo {
+        fn no_args() {}
+
+        fn with_return() -> &str {}
```
compiler-errors: For the record, `with_return` isn't valid Rust:

```
error[E0106]: missing lifetime specifier
 --> src/lib.rs:3:29
  |
3 |         fn with_return() -> &str {}
  |                             ^ expected named lifetime parameter
  |
  = help: this function's return type contains a borrowed value, but there is no value for it to be borrowed from
help: consider using the `'static` lifetime
  |
3 |         fn with_return() -> &'static str {}
```

tests/ui/fn/ok.🔬
```diff
@@ -0,0 +1,13 @@
+[
+    crate Foo {
+        fn no_args() {}
```
oli-obk:
```suggestion
        fn no_args<>() {}
```

I'd guess you need sth like this, but not sure how formality parses functions

tests/ui/fn/ok.🔬
```diff
@@ -0,0 +1,13 @@
...
+
+        fn with_return() -> String {}
```
oli-obk: Formality won't know this type, it has no standard library

tests/ui/fn/ok.🔬
```diff
@@ -0,0 +1,13 @@
...
+
+        fn is_true(v: bool) -> bool {}
+
+        fn add(x: i64, y: i64) -> i64 {}
+
+        fn return_multi(a: i32, b: i32) -> (i32, i32) {}
```
oli-obk: Not sure formality knows tuples

oli-obk (changes requested):

oli-obk:
I haven't forgotten about this 😆

tests/ui/fn/ok.stderr
```diff
@@ -0,0 +1 @@
+Error: failed to prove {@ wf((rigid &(shared) !lt_2 !ty_1))} given {}, got {}
```
oli-obk: about which function is this error?

maybe create one test per failing function.

tests/ui/fn/ok.🔬
```diff
@@ -0,0 +1,10 @@
+[
+    crate Foo {
+        fn no_args<>() -> () where [] { trusted }
+        fn one_args<ty T, lt l>(&l T) -> () where [] { trusted }
+        fn one_return<ty T, lt l, ty U>(&l T) -> U where [] { trusted }
```
oli-obk: Leave a comment explaining what Rust signature this signature is equivalent to. this is `fn foo<'a, T, U>(_: &'a T) -> U`, and previously you had different functions.

Also leave a comment explaining what the tests are testing, as I can't tell from the test.

nikomatsakis (approved):

PR #141, opened 2023-10-08
update SMIR dependency to latest nightly 

oli-obk:


nikomatsakis (approved):

spastorino:
Great, silly me because I had this change since a while in my local branch :).

PR #145, opened 2023-10-12
(Mostly) add AliasEq and start porting chalk projection tests

jackh726:
A decent start :)


tests/projection.rs
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+]";
+
+
+#[test]
+fn projection_equality() {
+    expect_test::expect![[r#"
+        Ok(
+            {},
```
jackh726: This isn't right :) Should be ?0 = u32. The first goal (the AliasEq) finds this constraint, but after solving the `S: Trait1<>` goal we don't have any

tests/projection.rs
```diff
@@ -0,0 +1,287 @@
...
+        )
+    "#]]
+    .assert_debug_eq(&test_where_clause(
+        PROJECTION_EQUALITY,
+        "exists<ty U> {} => { S: Trait1<>, <S as Trait1<>>::Type<> = U }",
```
jackh726: Removing the `S: Trait1<>` gives the expected constraints!

tests/projection.rs
compiler-errors: This is an outlives predicate

crates/formality-prove/src/prove/prove_wf.rs
```diff
@@ -39,10 +39,33 @@ judgment_fn! {
             (prove_wf(decls, env, assumptions, RigidTy { name: RigidName::ScalarId(_), parameters }) => c)
         )
 
+        (
+            (for_all(&decls, &env, &assumptions, &parameters, &prove_wf) => c)
+            --- ("integers and booleans")
```
nikomatsakis:
```suggestion
            --- ("ADT")
```

nikomatsakis: we should be checking the where-clauses on the ADT declaration, but let's leave that for a follow-up PR

nikomatsakis (approved):

PR #149, opened 2023-10-31
Make formality core reusable

nikomatsakis:
This branch extracts `formality_core` into a reusable abstraction so that one can reuse the formality system with other languages. I also copied in some of the docs I wrote and added a few more. It still needs a bunch of work before it's really usable. I'd like to make an example with a simple language and rework the docs to show that is built, piece by piece, but for now, this will do.


nikomatsakis:
Given that tests pass, I'm going to go ahead and merge this.

PR #151, opened 2023-11-01
Publish formality core

nikomatsakis:
Changes needed to publish formality-macros and formality-core to crates.io (done!). I want to be able to experiment with these for other languages. Eventually I'd like to publish all of a-mir-formality, just because.

I made the types team owners of these crates, but are there other owners I should add? I expect so, I didn't see obvious docs on forge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,24 @@
 version = 3
 
 [[package]]
+name = "a-mir-formality"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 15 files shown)


lqd:
Maybe the compiler team procedures could be useful as inspiration https://github.com/rust-lang/compiler-team/blob/master/content/procedures/crates.md but it hasn’t been updated in a long while and could be out-of-date.

PR #152, opened 2023-11-04
improve term macros

nikomatsakis:
I need to finish up the docs, but this branch makes the `#[term]` and parse macros much nicer. Various changes:

* Cleanup hygiene so that you don't have to bring things in scope to use the macros
* Permit you to use `#[term]` macro even if some of the arms contain variables, so long as they are properly annotated (folding variables is a bit tricky)
* Allow you to declare `#[customize]` items on terms, so that you can have custom parsing
* Cleanup the parsing to use a more structured API and to track reductions, so we can avoid some false conflicts
* Allow variants to declare precedence

This combination means that we can remove most of the custom parse/debug impls, and the ones that are left seem pretty reasonable overall. Plus the debug output is much nicer now, since we print things like `Vec<u32>` as `Vec<u32>` and not `(rigid (Vec) u32)`. 

There are still some flaws, most notably we don't pretty print associated types very well. I think the fix there is that we should actually track the number of parameters used for the trait vs the GAT in the `AssocTyId`. But that feels like something to manage in a separate PR.

To do before merging:

* [x] Document authoring of new parsers and new attributes

Follow-up:

* [x] Track GAT arity for associated types.


crates/formality-prove/src/test/eq_assumptions.rs
```diff
@@ -45,8 +45,8 @@ fn test_b() {
                 },
                 known_true: true,
                 substitution: {
-                    ?ty_1 => (rigid (adt Vec) (rigid (scalar u32))),
-                    ?ty_2 => (rigid (scalar u32)),
+                    ?ty_1 => Vec<u32>,
```
compiler-errors, line 48: very nice

PR #153, opened 2023-11-09
Improve parsing even further

nikomatsakis:
Parsing now supports:

* [x] Optional delimited vectors, so you don't have to write `Foo<>`
* [x] Guard clauses, so you don't have to write `struct Foo where [] {` but can  just write `struct Foo {`
* [x] Precedence and ambiguous grammars, so you can write `Expr = Expr + Expr` and things like that.

also adds a custom testing framework (`parser-torture-tests`) as well as an example language (`formality-eg`). The formality language is incomplete.


nikomatsakis:
I guess we didn't enable bors yet. :)

PR #155, opened 2023-11-11
Implement unsafe trait support (continued)

lqd:
This PR supersedes #128:
- rebases most of @yoshuawuyts's work over Niko's recent parser changes
- also integrates Niko's pretty-printer fix on optional fields, that way the spurious separators present in the previous PR are now gone
- matches rustc's error messages on safety mismatches
- adds E0198 check for negative impls
- adds a bunch of tests for the different combinations of safe/unsafe trait, safe/unsafe/negative impl
- optional: also adds `rustfmt` as a required component in the toolchain, just so it's easier to format things. And with that, removes some formatting weirdness that break `rustfmt`, and formats stray rogue files. I can remove these commits from the PR if you prefer not to have them.


crates/formality-prove/src/decls.rs
```diff
@@ -135,15 +140,26 @@ pub struct NegImplDeclBoundData {
     pub where_clause: Wcs,
 }
 
+#[term]
+#[derive(Default)]
+pub enum Safety {
+    #[default]
```
nikomatsakis, line 150: I didn't know you could do this, nifty.

crates/formality-prove/src/decls.rs
```diff
@@ -141,13 +143,23 @@ pub struct NegImplDeclBoundData {
...
 #[term]
+#[customize(debug)]
 #[derive(Default)]
 pub enum Safety {
     #[default]
     Safe,
     Unsafe,
 }
 
+impl fmt::Debug for Safety {
```
nikomatsakis, line 155: Isn't this the default behavior that `#[term]` would generate?

crates/formality-check/src/impls.rs
```diff
@@ -17,10 +17,8 @@ use formality_types::{
 };
 
 impl super::Check<'_> {
-    #[context("check_trait_impl({v:?})")]
-    pub(super) fn check_trait_impl(&self, v: &TraitImpl) -> Fallible<()> {
-        let TraitImpl { binder, safety } = v;
```
nikomatsakis: JFYI I use struct patterns a lot as a deliberate style choice -- I want compilation errors when/if add'l fields are added to the struct -- but in this case it doesn't seem very critical.

crates/formality-check/src/impls.rs
```diff
@@ -74,8 +72,8 @@ impl super::Check<'_> {
     }
 
     /// Validate that the declared safety of an impl matches the one from the trait declaration.
-    fn check_safety_matches(&self, trait_decl: &Trait, trait_impl: &Safety) -> Fallible<()> {
-        if trait_decl.safety != *trait_impl {
+    fn check_safety_matches(&self, trait_decl: &Trait, trait_impl: &TraitImpl) -> Fallible<()> {
```
nikomatsakis, line 81: ...and I like this version of the fn better

nikomatsakis (approved):

crates/formality-check/src/impls.rs
nikomatsakis: It's more for the result when opening `binder`

crates/formality-prove/src/decls.rs
lqd: Oh I didn't know this, or test that it would. I'll try, and open a PR to clean this up.

crates/formality-check/src/impls.rs
lqd: Make sense. I would add this back in this PR but it's already merged :) I'll open another PR.

crates/formality-prove/src/decls.rs
lqd: I've fixed this in #156

crates/formality-check/src/impls.rs
lqd: Even if it's not critical per se, I've re-introduced it in #156 here for consistency and easier evolvability, as well as to the negative impl counterpart.

PR #162, opened 2024-02-08
typo: fixed keyword on sample code

hackaugusto:
Just a small fix, the sample for the `enum` was using the wrong keyword.


oli-obk:
Thanks!

Issue #157, opened 2023-11-13
Obviously overlapping blanket impls are accepted


CheaterCodes:
In the hope of progressing https://github.com/rust-lang/rust/issues/20400, I stumbled across https://github.com/rust-lang/rfcs/pull/1672.
Given [this comment](https://github.com/rust-lang/rfcs/pull/1672#issuecomment-1748505018), I decided to properly investigate a-mir-formality.

As it turns out, the current overlap checks are very generous. I don't know whether this is intentional and I just misunderstood the purpose of this tool, or if it's an oversight, or simply incompleteness. In any case, the following test is currently accepted by a-mir-formality, and I personally think it should (obviously) not be:

```rust
[
    crate core {
        trait Foo { }

        trait Bar { }

        impl<ty T> Foo for T { }
        
        impl<ty T> Foo for T where T: Bar { }
    }
]
```

I also pushed this test `blanket_overlap` with the expected error output to a branch on my fork:
https://github.com/CheaterCodes/a-mir-formality/tree/overlap_test

It appears that currently, a-mir-formality is not considering that a downstream type may implement both of the above traits, instead it just determines that `may_be_remote` is provable false (which it isn't, if I understand this code correctly).

If my understanding so far is correct, I'll happily look into this more closely, but it seems like fairly major effort would be required.

compiler-errors:
For the record, `:` is outlives syntax. See https://github.com/rust-lang/a-mir-formality/issues/147.

I think the trait bound is expressed `Bar(T)`, though reference other tests to make sure.

lqd:
Coherence is also currently being worked on according to https://hackmd.io/gw4UQcN8Rm-MGqFFF0M6jA#Tracking-issue-for-a-mir-formality

CheaterCodes:
> For the record, `:` is outlives syntax. See #147.
> 
> I think the trait bound is expressed `Bar(T)`, though reference other tests to make sure.

I'm not sure if this is true. Other tests use the same syntax, and the grammar also suggests that this is correct:
https://github.com/rust-lang/a-mir-formality/blob/3468054f9c4d8c1c99013490c376a09886a2afc3/tests/ui/coherence_overlap/u32_T_where_T_Is_impls.%F0%9F%94%AC#L7

PR #167, opened 2024-05-13
cleanup and fix coherence rules

nikomatsakis:
this PR actually does two things that really ought not to be combined but I am lazy:

* move tests from using UI framework into test modules
* clean up the coherence rules

Fixes https://github.com/rust-lang/a-mir-formality/issues/157


compiler-errors:
r? lcnr

crates/formality-prove/src/prove/is_local.rs
```diff
@@ -37,24 +37,166 @@ use crate::{
...
+
+        (
+            // existential variables *could* be inferred to downstream types; depends on the substitution
+            // we ultimately have.
+            --- ("type variable")
+            (may_be_downstream_parameter(_decls, _env, _assumptions, TyData::Variable(Variable::ExistentialVar(_))) => ())
+        )
+
+        // If we can normalize `goal` to something else, and that normalized
+        // form may be downstream.
```
lcnr, line 108:
```
        // If `parameter` is an alias which refers a type which may be
        // from a downstream crate, it may be normalized to that type,
        // so recurse into it.
        //
        // We only do so if the alias cannot be normalized to a type which
        // is definitely not a downstream parameter.

```

lcnr:
woops, forgot i've added a nit

lcnr:
applied it directly

PR #171, opened 2024-05-14
continue coherence work

lcnr:
still a lot of TODOs, opening for review (and for me to be able to go to bed again)


crates/formality-prove/src/prove/is_local.rs
```diff
@@ -61,7 +62,7 @@ judgment_fn! {
         (
             // In principle this rule could be removed and preserve soundness,
             // but then we would accept code that is very prone to semver failures.
-            (may_not_be_provable(is_local_trait_ref(decls, env, assumptions, goal)) => ())
+            (may_not_be_provable(&env, |env| is_local_trait_ref(decls, &env, assumptions, goal)) => ())
             --- ("may be added by upstream in a minor release")
             (may_be_remote(decls, env, assumptions, goal) => ())
         )
```
lcnr: this is "could an upstream crate (a crate we know about) add an impl in a non breaking change"

crates/formality-prove/src/prove/is_local.rs
```diff
@@ -51,6 +51,7 @@ judgment_fn! {
         goal: TraitRef,
     ) => () {
         debug(assumptions, goal, decls, env)
+        assert(env.is_in_coherence_mode())
 
         (
             (may_be_downstream_trait_ref(decls, env, assumptions, goal) => ())
```
lcnr: where this instead asks "could this get added by a source we don't know about"

src/test/coherence_overlap.rs
```diff
@@ -275,3 +275,70 @@ fn T_and_Local_Bar_T() {
...
+                trait Foo<ty U> { }
+            },
+            crate foo {
+                struct LocalType {}
+                impl Foo<LocalType> for () {}
+
+                trait Overlap<ty U> {}
+                impl<ty T, ty U> Overlap<U> for T
+                where
+                    <T as Project>::Assoc: Foo<U> {}
```
lcnr, line 324: `<?t as Project>::Assoc: Foo<LocalType>` normalizes to `(): Foo<LocalType>`, but `<?t as Project>::Assoc` cannot be normalized in coherence mode

lcnr: need to figure out why this one still fails

nikomatsakis:
Will look at this soon!

PR #173, opened 2024-05-14
Some clippy fixes

lqd:
When adding rustfmt to CI, I noticed clippy was installed by default so I tried it and heard its many complaints with aplomb. (It hates the macro parsing in particular). 

This PR cuts down on the number of warnings it generated, in different commits so they can be dropped in case some of its suggestions are suboptimal.


compiler-errors (approved):

lqd:
(Thanks for the approval, I don't have merge rights tho)

PR #174, opened 2024-06-17
use underscore impls where we can

nikomatsakis:


lcnr:
aren't those the default in `synstructure` now? https://docs.rs/synstructure/latest/synstructure/struct.Structure.html#method.underscore_const

PR #175, opened 2024-06-22
track adt variant and field names do not conflict

shua:
The intent of this change is to add a check for adts that enum variant names are unique, and that struct and enum field names are unique (within a given enum variant).


crates/formality-check/src/adts.rs
```diff
@@ -15,10 +18,17 @@ impl super::Check<'_> {
 
         self.prove_where_clauses_well_formed(&env, &where_clauses, &where_clauses)?;
 
-        // FIXME: check names are unique or integers from 0..n
-
-        for Variant { name: _, fields } in &variants {
-            for Field { name: _, ty } in fields {
+        // names is used to check that there are no name conflicts
+        let mut names = HashSet::new();
```
lcnr, line 13: can you move the name check above any trait solving? i.e. right after the `instantiate_universally`?

nikomatsakis (changes requested):
this looks good modulo @lcnr's suggestion

crates/formality-check/src/adts.rs
shua: I needed some way to get the term out of the `Binder`. I guess it doesn't matter how it's instantiated for the purposes of unique naming, so I chose `Binder::open()` but maybe there's a simpler way with eg `Binder::instantiate`?

nikomatsakis: you can use the `Binder::peek` method, @shua

nikomatsakis (approved):

crates/formality-check/src/adts.rs
```diff
@@ -6,6 +9,20 @@ impl super::Check<'_> {
     pub(super) fn check_adt(&self, adt: &Adt) -> Fallible<()> {
         let Adt { id: _, binder } = adt;
 
+        // names is used to check that there are no name conflicts
+        let mut names = HashSet::new();
+        for Variant { name, fields } in &adt.binder.open().1.variants {
```
nikomatsakis:
```suggestion
        for Variant { name, fields } in &adt.binder.skip().variants {
```

nikomatsakis (changes requested):

crates/formality-check/src/adts.rs
shua: I saw you merged this change, but `skip` doesn't exist on binder so I changed it to the previously mentioned `peek` method instead.

nikomatsakis: Oh, ooops, thanks!

nikomatsakis (approved):

src/test/mod.rs
```diff
@@ -136,6 +136,41 @@ fn basic_where_clauses_fail() {
...
+        [
+            crate Foo {
+                enum Bar {
+                    Baz{},
+                    Baz{},
+                }
+            }
+        ]
+
+        [ /* TODO */ ]
```
nikomatsakis: add suggested messages in here

src/test/mod.rs
```diff
@@ -136,6 +136,41 @@ fn basic_where_clauses_fail() {
...
+        [
+            crate Foo {
+                struct Bar {
+                    baz: (),
+                    baz: (),
+                }
+            }
+        ]
+
+        [ /* TODO */ ]
```
nikomatsakis: and here

nikomatsakis (changes requested):
Almost there!

src/test/mod.rs
shua: Ah, I copied this from pre-existing tests in this file, so I had no idea what to put here, but scrolled down and found some other examples. This is supposed to be some summary of what's wrong with the code?

shua: ah wait, maybe instead it's some relevant substring in the error output that shouldn't change even when you update the expect output? (guessing from `assert_err!` naming that tt `must_have` and the other `expect`)

nikomatsakis (approved):

PR #177, opened 2024-06-30
replace custom wf check with well_formed constructor 

shua:
This change is mostly motivated by a comment on `prove_where_clause_well_formed`:
> FIXME(oli-obk): figure out why is this a function and not a `judgment_fn`.

I originally started with converting it to a `judgement_fn` but found I was duplicating a lot of logic from `prove_wc` which already exists. So I settled on moving the logic to a `WhereClause::well_formed` method which creates a set of goals to prove, and then calls `prove_wc` on those goals. 

I'm not _certain_ this is an improvement on the current code. There is more code being run (constructing a set of goals, then checking them is more than just checking), so I wouldn't expect any performance improvement, and it sort of looks like the same code just in a different place, so it's not removing much. The only justification I have is that I wouldn't ever do the reverse of this change (remove the `well_formed` constructor in favour of custom well_formed check).

### DIFF
--- a/crates/formality-check/src/where_clauses.rs
+++ b/crates/formality-check/src/where_clauses.rs
@@ -1,85 +1,20 @@
 use fn_error_context::context;
-use formality_core::Upcast;
 use formality_prove::Env;
-use formality_rust::{
-    grammar::{WhereClause, WhereClauseData},
-    prove::ToWcs,
-};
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):
I think it's good. I've been a bit reluctant to add "synthetic" predices if we can avoid it, but this feels like one we are going to want.

PR #178, opened 2024-06-30
inline PR into WcData

shua:
This change inlines the `PR` enum variants into the `WcData` enum. The comment on `PR` "we need to come up with a better name for this" indicates this type is not encoding some deeper logical idea. This change passing all tests indicates that the type is also not necessary.


nikomatsakis (approved):
makes sense. it used to be more important I think. or maybe it never was!

PR #179, opened 2024-07-01
Check trait items for duplicate function names or associated type names

FullyNonlinear:
This PR adds checks for duplicate function names or duplicate associated type names.

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -133,3 +133,48 @@ fn basic_where_clauses_fail() {
                                    expression evaluated to an empty collection: `decls.trait_invariants()`"#]]
     )
 }
+
+#[test]
+fn trait_items_with_duplicate_fn_names() {
+    crate::assert_err!(
+        [
+            crate core {
+                trait A {
+                    fn a() -> ();
+                    fn a() -> ();
+                }
+            }
+        ]
+
+        [ /* TODO */ ]
+
+        expect_test::expect![[r#"
+            check_trait(A)
+
+            Caused by:
+                the function name `a` is defined multiple times"#]]
+
+    );
+}
+
+#[test]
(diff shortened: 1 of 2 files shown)


src/test/mod.rs
```diff
@@ -133,3 +133,48 @@ fn basic_where_clauses_fail() {
...
+        [
+            crate core {
+                trait A {
+                    fn a() -> ();
+                    fn a() -> ();
+                }
+            }
+        ]
+
+        [ /* TODO */ ]
```
nikomatsakis:
```suggestion
        [ "function name `a` is defined multiple times" ]
```

try that, I think

src/test/mod.rs
```diff
@@ -133,3 +133,48 @@ fn basic_where_clauses_fail() {
...
+        [
+            crate core {
+                trait A {
+                    type Assoc : [];
+                    type Assoc : [];
+                }
+            }
+        ]
+
+        [ /* TODO */ ]
```
nikomatsakis:
```suggestion
        [ "associated type name `Assoc` is defined multiple times" ]
```

nikomatsakis (changes requested):

src/test/mod.rs
FullyNonlinear: Have updated accordingly on this commit:  1cff4c250d9d7acfe95102a96070724d01b88757. (An extra comma is needed for the `must_have` parameter.)

src/test/mod.rs
FullyNonlinear: Have updated accordingly in this commit: 1cff4c250d9d7acfe95102a96070724d01b88757. (An extra comma is needed for the `must_have` parameter.)

nikomatsakis (approved):

PR #181, opened 2024-07-02
add some wf checks for fn types

shua:
This change adds a syntax for specifying FnDefs and FnPtrs as type parameters, `fn name<T, U, ...>` and `fn(T, U,...) -> V` respectively. The `fn` prefix is used to disambiguate between adt identifiers. In addition, some relevant well-formed cases have been added.


crates/formality-prove/src/test/occurs_check.rs
```diff
@@ -20,13 +20,13 @@ fn decls() -> Decls {
 fn direct_cycle() {
     test_prove(decls(), term("exists<ty A> {} => {A = Vec<A>}")).assert_err(
             expect![[r#"
-                judgment `prove_wc_list { goal: {?ty_0 = Vec<?ty_0>}, assumptions: {}, env: Env { variables: [?ty_0], bias: Soundness }, decls: decls(222, [trait Foo <ty> ], [impl <ty> Foo(Vec<^ty0_0>)], [], [], [], [], {}, {}) }` failed at the following rule(s):
```
nikomatsakis: we should probably cleanup the `Env` Debug impl to e.g. suppress empty lists...

crates/formality-types/src/grammar/ty/debug_impls.rs
```diff
@@ -32,6 +32,25 @@ impl Debug for RigidTy {
                     write!(f, "()")
                 }
             }
+            RigidName::FnDef(name) => {
+                let parameters = PrettyParameters::new("<", ">", parameters);
+                write!(f, "{name:?}{parameters:?}",)
```
nikomatsakis: I'm surprised this is not `fn {name:?}`...?

nikomatsakis (changes requested):
Basically looks good — are there any tests for function WF though? Also, the debug impl didn't quite look like what I expected, which is to output `fn name<...>` and `fn(...)`. I think it will just emit `name<...>`and `(...)`?

src/test/functions.rs
```diff
@@ -39,13 +39,20 @@ fn lifetime() {
...
+            }
+        ]
 
-        expect_test::expect![[r#"
-            judgment `prove_wc_list { goal: {@ wf(&!lt_0 !ty_1)}, assumptions: {}, env: Env { variables: [!lt_0, !ty_1], bias: Soundness }, decls: decls(222, [], [], [], [], [], [], {}, {}) }` failed at the following rule(s):
-              the rule "some" failed at step #0 (src/file.rs:LL:CC) because
-                judgment `prove_wc { goal: @ wf(&!lt_0 !ty_1), assumptions: {}, env: Env { variables: [!lt_0, !ty_1], bias: Soundness }, decls: decls(222, [], [], [], [], [], [], {}, {}) }` failed at the following rule(s):
-                  the rule "parameter well formed" failed at step #0 (src/file.rs:LL:CC) because
-                    judgment had no applicable rules: `prove_wf { goal: &!lt_0 !ty_1, assumptions: {}, env: Env { variables: [!lt_0, !ty_1], bias: Soundness }, decls: decls(222, [], [], [], [], [], [], {}, {}) }`"#]]
+        expect_test::expect!["()"]
```
shua: @nikomatsakis this is a test for fn well-formed. I can also add some tests for catching poorly formed fns.

crates/formality-types/src/grammar/ty/debug_impls.rs
shua: hmm, fair point, I would also expect `fn {name}`

crates/formality-prove/src/test/occurs_check.rs
shua: #187 smaller decls output

lcnr (commented):
mind blocking this PR on https://github.com/rust-lang/a-mir-formality/pull/187#issuecomment-2231242428 reviewing it with most of the changes being an added `[]` to decls isn't great

shua:
good time for re-review @lcnr or @nikomatsakis ?

nikomatsakis:
Yes! I have been very absent from a-mir-formality but I'm looking to jump back in and figure out next steps soon-ish anyway, so it is a good time.

shua:
going to close this for now. It's getting a bit stale, and it's easy to make similar changes in the future so I don't know if there's much benefit keeping it open.

PR #186, opened 2024-07-10
Use unique names for rules in prove_wc

FullyNonlinear:
If the names of the rules passed to `judgment_fn!` are not unique, it will be more difficult to deduce immediately which exact rule is picked from the debug log. This PR makes the names in `prove_wc` unique. Regarding the names themselves, any suggestion other than the ones in this PR will be happily accepted.


lcnr (approved):

FullyNonlinear:
The change of the expected output is auto-updated by using `UPDATE_EXPECT=1`. In principle, the test cases will not be worse.

nikomatsakis (approved):

PR #187, opened 2024-07-11
reduce decls Debug noise

shua:
Decls debug rendering is a bit noisy, especially since decls does not
change, but is printed for every step that failed.  With this change,
decls is only printed in the `prove` judgement, with the expectation that
this is the main entrypoint for proofs.

### DIFF
--- a/crates/formality-prove/src/prove.rs
+++ b/crates/formality-prove/src/prove.rs
@@ -13,8 +13,9 @@ mod prove_wc_list;
 mod prove_wf;
 
 pub use constraints::Constraints;
+use formality_core::judgment::{FailedRule, TryIntoIter};
 use formality_core::visit::CoreVisit;
-use formality_core::{ProvenSet, Upcast};
+use formality_core::{set, ProvenSet, Upcast};
 use formality_types::grammar::Wcs;
 use tracing::Level;
 
@@ -53,7 +54,20 @@ pub fn prove(
 
     assert!(env.encloses(term_in));
 
-    let result_set = prove_wc_list(decls, &env, assumptions, goal);
+    struct ProveFailureLabel(String);
+    let label = ProveFailureLabel(format!(
+        "prove {{ goal: {goal:?}, assumptions: {assumptions:?}, env: {env:?}, decls: {decls:?} }}"
+    ));
+    impl std::fmt::Debug for ProveFailureLabel {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            f.write_str(&self.0)
+        }
+    }
+    let result_set =
+        match prove_wc_list(decls, &env, assumptions, goal).try_into_iter(|| "".to_string()) {
+            Ok(s) => ProvenSet::from_iter(s),
+            Err(e) => ProvenSet::failed_rules(label, set![FailedRule::new(e)]),
+        };
 
     tracing::debug!(?result_set);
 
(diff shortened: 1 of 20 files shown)


FullyNonlinear:
In my personal opinion, I prefer the original output format. The major benefit of the original format is that I can _easily_ _definitively_ map each field output to the name of the field, thanks to the list of field outputs having fixed length. 

That being said, I don't have a problem with skipping fields that are empty containers, as long as the name of the fields that are non-empty get displayed in the output. 

In addition, if the verbosity of the debug output of `decls` is annoying enough, positioning at the end of the line after any other useful information could also be a solution. (I understand that `decls` does usually change across the lines. )

shua:
Can you give me an example of an issue you had where it was helpful to see specifically which field in `Decls` the item was in?

FullyNonlinear:
> Can you give me an example of an issue you had where it was helpful to see specifically which field in `Decls` the item was in?

Here is one example that immediately comes to my mind:
```
test test::coherence_orphan::CoreTraitLocal_for_AliasToKnown_in_Foo ... ok
prove goal={}, assumptions={}, env=Env { variables: [], bias: Soundness }, decls=decls(222, [trait Foo <ty> ], [impl <ty> Foo(^ty0_0) where {Foo(^ty0_0)}], [impl ! Foo(u32)], [], [], [], {Foo}, {})
```
In this example, there are two fields that both start with the prefix `impl`, namely `impl_decls` and `neg_impl_decls`. For someone who knows well how to parse the new output format already, they would know to look ahead further to figure out which is the usual `impl_decls` and which is the `neg_imple_decls`. (I actually didn't know that until I mapped the output to the fields under the current format.) For beginners who do not have that information, it would be hard for them to know definitively which subsequences belong to which field. (Again, this is just my personal experience and opinion :blush: )

shua:
That's fair, so with this PR the output would look like
```
decls[ trait Foo <ty>, impl <ty> Foo(^ty0_0) where {Foo(^ty_0)}, impl ! Foo(u32) ]
```
and you're issue was that it wasn't clear you needed to call eg `decls.neg_impl_decls(foo_trait_id)` to get that value? That makes sense, I was coming from "why did my proof fail?" but not I hadn't really thought of it from the point of view of a dev needing to add a lookup of some id and not sure which field to use.

I'll try a slightly more verbose output, thanks for the feedback.

lcnr:
as a general question: should we mention the decls at all in the proof trace?

I am not that happy about using `..` for empty fields but also don't like the current impl without field names too much. Ideally we could just avoid that entirely :3

shua:
@lcnr the decls should only have to be output once in the trace of a proof, because they shouldn't change. So I guess we could just print them in some top-level function like `prove_goal` and don't print them anywhere else?

lcnr:
> @lcnr the decls should only have to be output once in the trace of a proof, because they shouldn't change. So I guess we could just print them in some top-level function like `prove_goal` and don't print them anywhere else?

yeah, exactly :+1:

lcnr (approved):
removing decls feels ncie. This now adds an additional level of indentation :thinking: I still think that's a fairly nice improvement

cc @nikomatsakis

crates/formality-prove/src/prove.rs
```diff
@@ -53,7 +54,20 @@ pub fn prove(
 
     assert!(env.encloses(term_in));
 
-    let result_set = prove_wc_list(decls, &env, assumptions, goal);
+    struct ProveFailureLabel(String);
```
nikomatsakis, line 57: Question: is it even worth including the "decls" here? I'm inclined to think "not really".

nikomatsakis (changes requested):

crates/formality-prove/src/prove.rs
shua: I think it's nice to see it at least once in proof trace on failure.

PR #190, opened 2025-02-02
When proving WellFormedTraitRef, require the parameters to be well-formed

### LINKED ISSUE #189: The checks for is-implemented where clauses are too lax
`formality-check` isn't rejecting an is-implemented where clause with nonsense in the implementing type or in a parameter to the trait.

For example it doesn't reject a where clause whose implementing type doesn't exist:

````
trait Trait1 {}
struct S where Nonex: Trait1 {}
````

or which gives the implementing type the wrong number of parameters:

````
trait Trait1 {}
struct S1 {}
struct S2<ty T> where S1<T>: Trait1 {
    dummy: T
}
````

It accepts nonsensical types in trait parameters too:

```
trait Trait1<ty T> {}
struct S where u8: Trait1<Nonex> {}
````

I see:

 - `formality-check`'s checks amount to proving `wc.well_formed()` for the single where clause

 - `wc.well_formed()` amounts to proving `WellFormedTraitRef` for (say) `Nonex: Trait1`

 - the "trait well formed" rule in `prove_wc()` considers only the where clauses of the `TraitRef`'s underlying trait, ignoring the `TraitRef`'s parameters.

mattheww:
Closes #189

Also adds a rule to `prove_wf()` saying that the `'static` lifetime counts as well-formed.

### DIFF
--- a/crates/formality-prove/src/prove/prove_wc.rs
+++ b/crates/formality-prove/src/prove/prove_wc.rs
@@ -4,6 +4,7 @@ use formality_types::grammar::{Predicate, Relation, Wc, WcData, Wcs};
 use crate::{
     decls::Decls,
     prove::{
+        combinators::for_all,
         env::{Bias, Env},
         is_local::{is_local_trait_ref, may_be_remote},
         prove,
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):

nikomatsakis:
This looks good to me!

PR #191, opened 2025-02-24
merge in parsing improvements

nikomatsakis:
These improvements to the parser have been living on my branch for a while. I introduced them when working with the https://dada-lang.org/ model.


nikomatsakis:
OK, I've validated that dada-model builds with this branch.

PR #194, opened 2025-03-18
WIP: const effect

tiif:
Opened a draft so it is easier to fetch and push to this branch.

### DIFF
--- a/tests/const_trait.rs
+++ b/tests/const_trait.rs
@@ -94,13 +94,29 @@ fn test_const_fn_with_runtime_effect() {
   }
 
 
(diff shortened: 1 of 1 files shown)


crates/formality-check/src/fns.rs
```diff
@@ -1,9 +1,9 @@
 use formality_prove::Env;
 use formality_rust::{
-    grammar::{Fn, FnBoundData},
+    grammar::{Fn, FnBody, FnBoundData, MaybeFnBody},
```
nikomatsakis, line 3: Ignore this, I'm making a test :)

PR #195, opened 2025-07-04
Typeck mir fn bodies

tiif:
This PR added some basic type checking for mir fn bodies.

### DIFF
--- a/crates/formality-check/src/impls.rs
+++ b/crates/formality-check/src/impls.rs
@@ -1,5 +1,6 @@
 use anyhow::bail;
 
+use crate::CrateItem;
 use fn_error_context::context;
 use formality_core::Downcasted;
 use formality_prove::{Env, Safety};
@@ -18,7 +19,11 @@ use formality_types::{
 
 impl super::Check<'_> {
     #[context("check_trait_impl({trait_impl:?})")]
-    pub(super) fn check_trait_impl(&self, trait_impl: &TraitImpl) -> Fallible<()> {
+    pub(super) fn check_trait_impl(
+        &self,
+        trait_impl: &TraitImpl,
+        all_fn: &Vec<CrateItem>,
+    ) -> Fallible<()> {
         let TraitImpl { binder, safety: _ } = trait_impl;
 
         let mut env = Env::default();
@@ -48,7 +53,7 @@ impl super::Check<'_> {
         self.check_safety_matches(trait_decl, trait_impl)?;
 
         for impl_item in &impl_items {
-            self.check_trait_impl_item(&env, &where_clauses, &trait_items, impl_item)?;
+            self.check_trait_impl_item(&env, &where_clauses, &trait_items, impl_item, all_fn)?;
         }
 
         Ok(())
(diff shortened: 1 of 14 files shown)


src/test/mir_fn_bodies.rs
```diff
@@ -0,0 +1,100 @@
...
+        [
+            crate Foo {
+                fn foo (u32) -> u32 = minirust(v1) -> v0 {
+                    let v0: u32;
+                    let v1: u32;
+                    
+                    bb0: {
+                        v1 = v0;
+                    }
+                    return;
```
tiif, line 16: The parser does not accept any statement inside the basic block. So currently the valid syntax is only 
```
bb0: {}
```

crates/formality-check/src/impls.rs
```diff
@@ -18,7 +19,11 @@ use formality_types::{
 
 impl super::Check<'_> {
     #[context("check_trait_impl({trait_impl:?})")]
-    pub(super) fn check_trait_impl(&self, trait_impl: &TraitImpl) -> Fallible<()> {
+    pub(super) fn check_trait_impl(
+        &self,
+        trait_impl: &TraitImpl,
+        all_fn: &Vec<CrateItem>,
```
nikomatsakis: As we discussed live, we can use `self.program.items_from_all_crates()` instead.

crates/formality-prove/src/prove/prove_wc.rs
```diff
@@ -53,6 +54,14 @@ judgment_fn! {
             (prove_wc(decls, env, assumptions, WcData::Relation(goal)) => c)
         )
 
+        (
+            (if let Ty(_) = param1.clone())!
+            (if let Ty(_) = param2.clone())!
+            (if param1 == param2)!
+                ----------------------------- ("subtype - reflexive")
+            (prove_wc(_decls, env, _assumptions, WcData::Relation(Relation::Sub(param1, param2))) => Constraints::none(env))
```
nikomatsakis, line 63: Note to self: we need to do better here

crates/formality-check/src/impls.rs
tiif: This is fixed in https://github.com/rust-lang/a-mir-formality/pull/195/commits/761a275401b210317f1e71a5ad32996902c58629. In that commit, I passed CrateId information down because I think we need it to find the correct crate in the program.

tiif:
I briefly talked about supporting constant in ValueExpression, https://github.com/rust-lang/a-mir-formality/pull/195/commits/6ef7fa0dbe71176a86bc71eb2bb6ba4fdf49f348 is how it looks like for now. Evidently there are quite a lot of boilerplate code, and we probably want a more general version of const. 

I can strip that commit out if it is not desirable to land that for now.

crates/formality-rust/src/grammar/minirust.rs
```diff
@@ -122,13 +148,86 @@ pub enum ValueExpression {
...
+        | ScalarId::U64
+        | ScalarId::I8
+        | ScalarId::I16
+        | ScalarId::I32
+        | ScalarId::I64
+        | ScalarId::Usize
+        | ScalarId::Isize => true,
+        ScalarId::Bool => false,
+    }
+}
```
tiif, line 223: I think this function should live somewhere else, or should be done in a better way

crates/formality-core/src/fold.rs
tiif, line 1: It'd be nice improvement to add these manually, we need these for ``ConstTypePair`` in minirust body.

tiif (commented):
I slightly cleaned up the commit history, and pointed out things that I find iffy.

nikomatsakis (approved):

PR #196, opened 2025-07-15
Add checks for undeclared local_id at function argument and return position

tiif:


tiif:
This is an accident :D

PR #200, opened 2025-08-12
Refactor `ty_is_int`

### LINKED ISSUE #198: Refactor `ty_is_int`
Per tiif's commen t here, this function should be refactored as a judgment perhaps, or else move elsewhere:

_Originally posted by @tiif in https://github.com/rust-lang/a-mir-formality/pull/195#discussion_r2239851687_
            

This issue has been assigned to @tiif via [this comment](https://github.com/rust-lang/a-mir-formality/issues/198#issuecomment-3144563422).

tiif:
Ignore the first commit, it is just UPDATE_EXPECT=1 for the change in main branch

Fixed #198

### DIFF
--- a/crates/formality-check/src/mini_rust_check.rs
+++ b/crates/formality-check/src/mini_rust_check.rs
@@ -385,7 +385,7 @@ judgment_fn! {
         // For the rest of the case, it should fail.
 
         (
-            (prove_normalize(&decl, &env, &assumptions, ty) => (c1, p))
+            (prove_normalize(&decl, &env, &assumptions, ty) => (c1, p))!
             (let assumptions = c1.substitution().apply(&assumptions))
             (ty_is_int(&decl, &env, assumptions, p) => c2)
             ----------------------------- ("alias_ty is int")


crates/formality-check/src/mini_rust_check.rs
```diff
@@ -229,9 +228,7 @@ impl Check<'_> {
                 // Check if the value is well-formed.
                 let value_ty = self.check_value(typeck_env, switch_value).unwrap();
 
-                if !ty_is_int(value_ty) {
-                    bail!("The value used for switch must be an int.");
-                }
+                self.prove_goal(&typeck_env.env, Wcs::default(), value_ty.is_int())?;
```
lcnr: I personally don't like introducing a new judgement for this. I think if we want this check to be a proper judgement maybe add a *builtin trait* and prove `value_ty: BuiltinInteger` here :thinking: 

Even if it's a proper judegement:tm: but instead be just something else you can prove. Predicates feel *core* to the type system in a way in which "is this an integer" does not. That's just my personal vibe here though

What's the reasoning behind changing an assertion to an actual goal?

crates/formality-prove/src/prove/prove_wc.rs
```diff
@@ -145,5 +147,30 @@ judgment_fn! {
             ----------------------------- ("const has ty")
             (prove_wc(decls, env, assumptions, Predicate::ConstHasType(ct, ty)) => c)
         )
+
+        (
+            (is_int(decl, env, assumptions, ty) => c)
+            ----------------------------- ("ty is int")
+            (prove_wc(decl, env, assumptions, Relation::IsInt(ty)) => c)
```
lcnr: how does a-mir-formality handle normalization when proving predicates/relations?

Because right now `<u32 as Id>::This` fails to satisfy `IsInt`

crates/formality-check/src/mini_rust_check.rs
tiif: The reason of changing this to a goal is I feel we should have a way to say "this type is an int" somewhere instead of having this function https://github.com/rust-lang/a-mir-formality/pull/200/files#diff-b58482e54c5aa3c63282fc2a3a0aa88194e826004773bd7bc9e385f29b5c3e0d

But I do agree that adding a predicate for this might be an overkill.

crates/formality-prove/src/prove/prove_wc.rs
tiif: Yeah that's true, I did an experiment in https://github.com/rust-lang/a-mir-formality/pull/200/commits/331c7371a0de18dd4a8a1c63d218d2acb94ae216#diff-2ca1246445028c80247cd13d93fd973d22ca55cfb7a76c0903a04c09055d706c and it indeed failed. 

The place where normalization is done should be here https://github.com/rust-lang/a-mir-formality/blob/8b130f7816b57495f6d37f5dd4ef37c858d255f6/crates/formality-prove/src/prove/prove_normalize.rs#L17
It is possible to use ``prove_normalize`` when proving a goal, for example: https://github.com/rust-lang/a-mir-formality/blob/8b130f7816b57495f6d37f5dd4ef37c858d255f6/crates/formality-prove/src/prove/is_local.rs#L288-L295

tiif: But with the change in https://github.com/rust-lang/a-mir-formality/pull/200/commits/c0fbd8b3c3dd66602c3792929949164eba17f5a2, it is possible to make ``IsInt`` work with normalization.

crates/formality-check/src/mini_rust_check.rs
lcnr: could we instead just use a "non-predicate" judgement here. Same as we do for `trait_ref_is_knowable` or however it's called?

crates/formality-prove/src/prove/prove_wc.rs
```diff
@@ -145,5 +148,39 @@ judgment_fn! {
...
+            ----------------------------- ("rigidty is int")
+            (prove_wc(decl, env, assumptions, Relation::IsInt(ty)) => c)
+        )
+
+        (
+            (prove_normalize(&decl, &env, &assumptions, ty) => (c1, p))
+            (let assumptions = c1.substitution().apply(&assumptions))
+            (is_int(&decl, &env, assumptions, p) => c2)
+            ----------------------------- ("ty is int")
+            (prove_wc(decl, env, assumptions, Relation::IsInt(ty)) => c2)
```
lcnr: hmm, maybe also as a question for @nikomatsakis 

This currently means there are 2 leaf judegements for proving `is_int`, one with normalization and one without.

What does `prove_normalize` actually do :sweat_smile: is it "normalize until rigid [but at least 1 step]" or "normalize 1 step"?

I think regardless, I personally prefer the setup of `prove_wc(Relation::IsInt(ty)) -> normalize(ty) -> prove_wc(Relation::IsInt(normalized_ty))` so that there's only one "leaf" judgement here

crates/formality-check/src/mini_rust_check.rs
tiif: Oh, I forgot to mention that I was using a ``Relation`` instead of ``Predicate``.

From my understanding, we can only use a ``judgment_fn!`` through ``prove_wc``, and when we call ``prove_wc``, we need to pick one of the below as the Wc type:
```rust
pub enum WcData {
    /// Means the built-in relation holds.
    #[cast]
    Relation(Relation),

    /// Means the predicate holds.
    #[cast]
    Predicate(Predicate),

    // Equivalent to `for<'a>` except that it can also express `for<ty T>` and so forth:
    // means `$v0` is true for any value of the bound variables (e.g., `'a` or `T`).
    #[grammar(for $v0)]
    ForAll(Binder<Wc>),

    #[grammar(if $v0 $v1)]
    Implies(Wcs, Wc),
}
```
``Relation`` looks the most reasonable to me among all of them.

crates/formality-prove/src/prove/prove_wc.rs
tiif: I write two judgements because ``prove_normalize`` will fail if we pass an int type to it. I slightly changed the rule for ``ty_is_int`` in https://github.com/rust-lang/a-mir-formality/pull/200/commits/1f601cf38bab93b18a5bc3f488815aacabfba6e8 so it always normalize the type to rigid if possible, and if it can't be normalize to rigid, it should just fail.

nikomatsakis: Answer: `prove_normalize` does a single step. I think this logic is good, because the base mir-formality infra should handle cycles -- we should add a test for fun, though.

crates/formality-check/src/mini_rust_check.rs
```diff
@@ -374,3 +371,34 @@ struct TypeckEnv {
...
+        assumptions: Wcs,
+        ty: Parameter,
+    ) => Constraints {
+        debug(assumptions, ty, env)
+        // If the type that we are currently checking is rigid, check if it is an int.
+        // If the type can be normalized, normalize until rigid then check if it is an int.
+        // For the rest of the case, it should fail.
+
+        (
+            (prove_normalize(&decl, &env, &assumptions, ty) => (c1, p))
```
lcnr:
```suggestion
            (prove_normalize(&decl, &env, &assumptions, ty) => (c1, p))!
```
I'd guess we don't care about this candidate if stuff is unnormalizeable? or well, we currently use this judegement even for non alias types :thinking:  I feel like maybe limiting `prove_normalize` to require an `AliasTerm` (or whatever we've got in a-mir-formality) would make avoid this?

nikomatsakis: Yes, I debated about that. I think this is a good addition, the only downside would be if you *expect* it to be normalizable and it's not... you won't readily know why... might be that we can do better with these annotations.

lcnr: > I feel like maybe limiting `prove_normalize` to require an `AliasTerm` (or whatever we've got in a-mir-formality) would make avoid this?

Thinking about this, this seems generally desirable esp if it's literally *shallow one-step normalization*, using it for non-alias types seems weird imo

clearly not something that has to happen in this PR :grin:

tiif: Created an issue for that https://github.com/rust-lang/a-mir-formality/issues/201

Issue #201, opened 2025-08-20
Only pass normalizable type to ``prove_normalize``


tiif:
As per the comment here: https://github.com/rust-lang/a-mir-formality/pull/200#discussion_r2281486848

We should also rename ``prove_normalize`` to ``normalize_alias`` at the same time.

PR #203, opened 2025-10-08
Add documentation

tiif:
Cherry-picked documentation from https://github.com/rust-lang/a-mir-formality/pull/194


crates/formality-check/src/impls.rs
```diff
@@ -120,6 +120,44 @@ impl super::Check<'_> {
         }
     }
 
+    /// Check that function `ii_fn` that appears in an impl is valid.
+    /// This includes the core check from [`Self::check_fn`][],
```
lcnr:
```suggestion
    /// This includes the core check from [`Self::check_fn`],
```

crates/formality-check/src/impls.rs
```diff
@@ -120,6 +120,44 @@ impl super::Check<'_> {
...
+    /// }
+    ///
+    /// struct Cup<L>;
+    /// impl<L> Potable for Cup<L> // <-- env has `L` in scope
+    /// where
+    ///     L: Potable, // <-- `impl_assumptions`
+    /// {
+    ///     fn drink(&self) {} // <-- `ii_fn`
+    ///     fn smell(&self) {} // not currently being checked
+    /// }
```
lcnr, line 153: this example is not fully general. I think another important part is that the *instantiated* trait method where-clauses imply the impl method where-clauses

crates/formality-prove/src/prove/prove_eq.rs
```diff
@@ -64,8 +64,21 @@ judgment_fn! {
             (prove_eq(decls, env, assumptions, Variable::ExistentialVar(v), r) => c)
         )
 
+        // Example: we are trying to prove `x` (which equals `<SomeX<?V> as Iterator>::Item`)
+        // is equal to some type `z`.
         (
+            // Normalize `x` will find alternative "spellings" that it is equivalent to.
```
lcnr: do we use "spellings" somewhere else? 🤔 would otherwise like to avoid even more terminology here '^^

We do use "underlying type" in rustc I think. i don't remember whether we'e talked about that before, but I feel like maybe we should change `prove_normalize` to take an alias type and limit the candidate to only `x` which are aliases

tiif: :+1:  i changed the wording

and yeah we have an issue open for that https://github.com/rust-lang/a-mir-formality/issues/201, i planned to get to that soon.

PR #205, opened 2025-10-28
Support tuple in mir fn body

tiif:


crates/formality-check/src/mini_rust_check.rs
```diff
@@ -432,6 +421,47 @@ impl Check<'_> {
...
+                    bail!("The length of tuple declared is {declared_tuple_length}, the number of value provided is {value_length}");
+                }
+
+                let mut value_tys: Vec<Ty> = Vec::new();
+
+                for value_expression in value_expressions {
+                    // FIXME: we only support const in value expression of tuple for now, we can add support
+                    // more in future.
+                    let Constant(_) = value_expression else {
+                        bail!("Only Constant is supported in ValueExpression::Struct for now.")
```
lcnr, line 448:
```suggestion
                        bail!("Only Constant is supported in ValueExpression::Tuple for now.")
```

crates/formality-check/src/mini_rust_check.rs
```diff
@@ -281,33 +281,22 @@ impl Check<'_> {
...
-                ) = self.decls.adt_decl(&adt_id).binder.open();
-                let AdtDeclVariant { name, fields } = variants.last().unwrap();
 
-                if *name != VariantId::for_struct() {
-                    bail!("The local used for field projection must be struct.")
+                if !ty.is_tuple() && !ty.is_adt() {
+                    bail!("The local used for field projection must be ADT or Tuple")
                 }
 
-                // Check if the index is valid for the tuple.
```
tiif, line 305: This is wrong, While tuple ``(i32, u32)``'s parameter is ``[i32, u32]``,
```rust
struct Foo<T> {
    value: T
}
```
struct ``Foo``'s parameter is T. 

But I don't like how we need to go through ``self.decls`` to get the struct field information.

PR #207, opened 2025-12-10
Cleanup and extend tests 

nikomatsakis:
Work done with @lqd and @tiif today: added support for `&mut`, wrote some new tests (including one FIXME), and added a test for NLL Problem Case #3 (which behaved as expected).

r? @jackh726

### DIFF
--- a/crates/formality-rust/src/grammar/minirust.rs
+++ b/crates/formality-rust/src/grammar/minirust.rs
@@ -144,6 +144,12 @@ pub enum Statement {
 /// Based on [MiniRust terminators](https://github.com/minirust/minirust/blob/9ae11cc202d040f08bc13ec5254d3d41d5f3cc25/spec/lang/syntax.md#statements-terminators).
 #[term]
 pub enum Terminator {
+    // Nondeterminstically jump to one of the given blocks.
+    //
+    // In practice, the compiler "is very very likely" to always
+    // select the first one. But the static analysis does not rely on that.
+    //
+    // NB: Diverges from MiniRust
     #[grammar(goto $,v0)]
     Goto(Vec<BbId>),
 


src/test/mir_fn_bodies.rs
```diff
@@ -867,6 +1055,109 @@ fn test_ref_not_subtype() {
...
+                        }
+                    }
+                };
+            }
+        ]
+    )
+}
+
+/// Upcasting from `'a` to `'c` should be allowed because of
+/// the transitive relationship.
```
lqd: No biggie, but since the doc comment is a copypasta, maybe something like the following would work?

```suggestion
/// Upcasting from `'a` to `'c` errors because of a missing
/// declared relationship to complete the transitive chain.
```

We can fix it later ofc.

src/test/mir_fn_bodies.rs
```diff
@@ -826,18 +827,203 @@ fn test_borrow_check_basic() {
...
+                      => false, LtData::Variable(Variable::ExistentialVar(_)) => true,
+                      LtData::Variable(Variable::BoundVar(_)) =>
+                      panic!("cannot outlive a bound var"),
+                  }, Parameter::Const(_) => panic!("cannot outlive a constant"),
+              })`"#]]
+    )
+}
+
+/// Test the holding a shared reference to a local
+/// integer variable prevents it from being incremented.
```
jackh726: This comment looks wrong.

src/test/mir_fn_bodies.rs
```diff
@@ -826,18 +827,203 @@ fn test_borrow_check_basic() {
...
+///
+/// The test is equivalent to:
+/// ```rust,ignore
+/// fn min_problem_case_3(m: &mut Map) -> &mut Map {
+///     let n: &mut Map = &mut *m;
+///     if some_condition() {
+///     }
+///     let o: &mut Map = &mut *m;
+///     return o;
+/// }
```
jackh726, line 978: I thought "problem 3" was something that doesn't compile today but should (also, as an aside, we should have this example in some more accessible place to refer back to). But, I don't see why this doesn't compile today.

crates/formality-rust/src/grammar/minirust.rs
```diff
@@ -144,8 +144,8 @@ pub enum Statement {
 /// Based on [MiniRust terminators](https://github.com/minirust/minirust/blob/9ae11cc202d040f08bc13ec5254d3d41d5f3cc25/spec/lang/syntax.md#statements-terminators).
 #[term]
 pub enum Terminator {
-    #[grammar(goto $v0)]
-    Goto(BbId),
+    #[grammar(goto $,v0)]
+    Goto(Vec<BbId>),
```
jackh726, line 154: We should leave a comment that this diverges from MiniRust (and why) and what the Vec represents.

src/test/mir_fn_bodies.rs
lqd: This test is the problem 3 test with the erroneous stmt removed ("min", but like, "too minimal" below). It compiles today, and formality accepts it. Apart from the naming that could be improved for clarity, it's just the dual that passes, and showcases/"ensures" that the implementation is actually doing something when it's computing the previous error.

src/test/mir_fn_bodies.rs
lqd: It's what I was referring to [here](https://github.com/rust-lang/a-mir-formality/pull/207#issuecomment-3637664290)

crates/formality-rust/src/grammar/minirust.rs
lqd: some of that explanation went into the commit mesg yeah, https://github.com/rust-lang/a-mir-formality/pull/207/commits/5269743d79a0fbd0a2af7e16564eaf128a7790a6, we can pull some of it here

PR #210, opened 2025-12-17
Accumulate outlives separately, read-only typeck-env

nikomatsakis:
Per our discussion today, we are moving towards modeling polonius-α which will require a context-dependent notion of outlives. This will also imply merging the type-check + borrow-check rules. These refactorings move us closer to that world.

r? @lqd or @jackh726

### DIFF
--- a/crates/formality-check/src/lib.rs
+++ b/crates/formality-check/src/lib.rs
@@ -130,6 +130,7 @@ impl Check<'_> {
             CrateItem::NegTraitImpl(i) => self.check_neg_trait_impl(i),
             CrateItem::Test(t) => self.check_test(t),
             CrateItem::FeatureGate(_feature_gate) => {
+                // FIXME(#212): reject duplicate feature gates within a crate
                 Ok(ProofTree::leaf("feature gates are OK with me!"))
             }
         }


jackh726 (approved):

crates/formality-check/src/borrow_check/nll.rs
```diff
@@ -134,14 +134,18 @@ enum AccessKind {
 /// Given the `TypeckEnv`, which includes a (populated) list of `pending_outlives`
```
lqd: As we've extracted the outlives out of the env, this doc comment was slightly invalidated, so maybe:

```suggestion
/// Given the `TypeckEnv`, and a (populated) list of `pending_outlives`
```

crates/formality-check/src/lib.rs
```diff
@@ -128,6 +129,9 @@ impl Check<'_> {
             CrateItem::Fn(f) => self.check_free_fn(f, crate_id),
             CrateItem::NegTraitImpl(i) => self.check_neg_trait_impl(i),
             CrateItem::Test(t) => self.check_test(t),
+            CrateItem::FeatureGate(_feature_gate) => {
+                Ok(ProofTree::leaf("feature gates are OK with me!"))
```
lqd, line 134: maybe we can add a comment like "TODO: reject duplicate gates like rustc"

src/test/mir_fn_bodies.rs
```diff
@@ -488,9 +488,10 @@ fn test_function_arg_number_mismatch() {
 }
 
 // Test the behaviour of having unitialised return local variable.
+// FIXME(#209): This test should fail but currently passes due to removed return place initialization check
```
lqd, line 491: we could maybe just `#[ignore]` that test with the comment, rather than changing the assertion from err to ok?

nikomatsakis: I generally prefer to have the test assert the wrong behavior when possible, so you don't forget to re-enable it

crates/formality-check/src/lib.rs
nikomatsakis: filed #212

nikomatsakis:
```suggestion
                // FIXME(#212): reject duplicate feature gates within a crate
                Ok(ProofTree::leaf("feature gates are OK with me!"))
```

PR #225, opened 2026-01-14
Add issue number for FIXME or remove it

tiif:


crates/formality-check/src/borrow_check/nll.rs
```diff
@@ -959,48 +955,6 @@ fn transitively_outlived_by(
...
-            CoreVariable::ExistentialVar(_) => {
-                panic!("do not expect existentials in borrow checker")
-            }
-
-            // e.g., the `'a' in `for<'a> fn(&'a u32)` -- this we can ignore because they don't represent a live borrow
-            CoreVariable::BoundVar(_) => set![],
-        },
-    }
-}
-
```
tiif, line 1003: we don't use this anywhere, not sure why there is no unused warning for this

src/test/borrowck.rs
tiif, line 388: Left this here because we already fixed this in ``living-large``

crates/formality-prove/src/decls.rs
```diff
@@ -282,7 +282,7 @@ impl AliasBoundDecl {
 #[term($alias : $ensures $:where $where_clause)]
 pub struct AliasBoundDeclBoundData {
     pub alias: AliasTy,
-    // FIXME: this is currently encoded as something like `<T> [T: Foo]` where
+    // This is currently encoded as something like `<T> [T: Foo]` where
```
tiif: no concrete idea on how we would want to fix this, so remove the FIXME, guess we will eventually revisit this if this is not ok.

crates/formality-check/src/traits.rs
```diff
@@ -116,7 +116,7 @@ impl super::Check<'_> {
             &where_clauses,
         )?;
 
-        // FIXME: Do we prove ensures WF? And what do we assume when we do so?
+        // Do we prove ensures WF? And what do we assume when we do so?
```
tiif: this feels more like a question that we should be aware of instead of FIXME

crates/formality-prove/src/decls.rs
lqd: so we can keep the fixme for now?

crates/formality-check/src/traits.rs
lqd: Maybe we should either answer it and remove the comment altogether, or just keep the fixme to not forget about it, or have an issue?

crates/formality-check/src/mini_rust_check.rs
```diff
@@ -95,7 +94,7 @@ impl Check<'_> {
         }
 
         let mut env = TypeckEnv {
-            program: Arc::new(self.program.clone()), // FIXME
```
lqd, line 98: what was this one about, cloning the program?

crates/formality-prove/src/prove/negation.rs
```diff
@@ -35,7 +35,7 @@ pub fn is_definitely_not_proveable<T: CoreFold<FormalityLang>>(
     // where `not (T: Foo)`, only that we could not prove it is true for all `T`
     // (put another way, it may still be true for all T, we just couldn't prove it).
     //
-    // (FIXME(nikomatsakis): This sounds suspicious, if we *truly* handle bias correctly,
+    // (notes from nikomatsakis): This sounds suspicious, if we *truly* handle bias correctly,
```
lqd: should we have an issue for this one?

crates/formality-prove/src/prove/is_local.rs
```diff
@@ -296,7 +296,7 @@ judgment_fn! {
         // Fundamental types are local if all their arguments are local.
         (
             (if is_fundamental(&decls, &name))
-            (for_all(&decls, &env, &assumptions, &parameters, &is_local_parameter) => c) // FIXME: should be `is_local_parameter`
```
lqd, line 299: was this about the reference?

lqd:
Apart from the couple nits, this LGTM. Thanks for doing this cleanup!

We should discuss in today's meeting.

crates/formality-check/src/traits.rs
tiif: i think it'd be nice to discuss this in today's meeting and decide what to do with it

crates/formality-check/src/mini_rust_check.rs
tiif: I couldn't remember what this is about, but I suppose it is probably about "it'd be nice if we don't put entire program in typeck env", but i don't see a way of not doing this yet and this doesn't feel too bad when i take a look at it now :thinking:

crates/formality-prove/src/decls.rs
tiif: yup that's alright too

crates/formality-prove/src/prove/negation.rs
tiif: yup sure

crates/formality-prove/src/prove/is_local.rs
tiif: it doesn't really look like it, we could ask niko later.

crates/formality-prove/src/decls.rs
lqd: So we want to add a link to https://github.com/rust-lang/a-mir-formality/issues/226 to this fixme, right?

tiif: yup, pushed the commit that contains the issue number.

crates/formality-prove/src/prove/negation.rs
lqd: https://github.com/rust-lang/a-mir-formality/issues/227

crates/formality-check/src/mini_rust_check.rs
nikomatsakis: yeah whatever it's fine

crates/formality-check/src/traits.rs
nikomatsakis: The question is related to whether this is an error...

```rust
trait Foo {
    type Bar: Baz<String>;
}

trait Baz<T: Copy> { }
```

...seems like it *should* be, or at least a warning, but we aren't enforcing that, and that's partly entangled with implied bounds questions. i.e., Maybe the trait is ok, it can just never be implemented?

Seems related to so-called "trivial bounds"... e.g., I'd like to accept this...

```rust
fn foo()
where
    String: Copy,
{
    ...
}
```

...so ... how does that relate here?

nikomatsakis:
```suggestion
        // FIXME(#228) Do we prove ensures WF? And what do we assume when we do so?
```

nikomatsakis (approved):

PR #237, opened 2026-01-22
introduce `a in foo` form

nikomatsakis:
Previously I overloaded (a => b) where `a` could be any iterable. But in explaining this to @BoxyUwU recently I thought it'd be nicer to have the iterable case separated out from other judgments.

What do people think?

### DIFF
--- a/crates/formality-core/src/judgment/test_for_all.rs
+++ b/crates/formality-core/src/judgment/test_for_all.rs
@@ -42,7 +42,7 @@ fn test_for_all_failure() {
     let nums = vec![Num(2), Num(3), Num(6)];
     all_even(nums).assert_err(expect_test::expect![[r#"
         the rule "all_even" at (test_for_all.rs) failed because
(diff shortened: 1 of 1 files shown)


crates/formality-core/src/judgment/test_filtered.rs
```diff
@@ -26,7 +26,7 @@ judgment_fn!(
         debug(node, g)
 
         (
-            (graph.successors(a) => b)
+            (b in graph.successors(a))
```
nikomatsakis, line 29: I was wondering if `(for b in graph.successors(a))` might be nice--- I was thinking that it was analogous to

$$
b \in graph.successors(a)
$$

in Mathematical notation, but "for some `b` in `graph.successors(a)`" reads more English-y

tiif: Without knowing we have ``for_all``, ``for b in graph.successors(a)`` can also be interpreted as "for all ``b`` in ``graph.successors(a)``", imo.

PR #238, opened 2026-01-28
Generically constant

nikomatsakis:
Begin working with @BoxyUwU to update the const generics model.

r? @BoxyUwU or perhaps @oli-obk

### DIFF
--- a/crates/formality-types/src/grammar/consts.rs
+++ b/crates/formality-types/src/grammar/consts.rs
@@ -1,7 +1,7 @@
 use crate::grammar::{ParameterKind, Parameters, RigidName, ScalarId};
 
 use super::{Parameter, Variable};
-use formality_core::{DowncastTo, Upcast, UpcastFrom, cast_impl, term};
+use formality_core::{cast_impl, term, DowncastTo, Upcast, UpcastFrom};
 use std::sync::Arc;
 
 #[term]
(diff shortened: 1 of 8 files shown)


crates/formality-types/src/grammar/consts.rs
```diff
@@ -22,34 +21,79 @@ impl Const {
...
+    I32(i32),
+    #[grammar(i64($v0))]
+    I64(i64),
+    #[grammar($v0)]
+    Bool(bool),
+    #[grammar(usize($v0))]
+    Usize(usize),
+    #[grammar(isize($v0))]
+    Isize(isize),
+}
```
nikomatsakis, line 70: Two notes:

(1) we are missing `char`
(2) it'd be nice to do a custom parse so we could accept e.g. `0_u8`

crates/formality-types/src/grammar/ty/parse_impls.rs
```diff
@@ -117,22 +115,3 @@ fn parse_parameters<'t>(
...
-            parser.parse_variant_variable("Variable", Precedence::default(), ParameterKind::Const);
-
-            parser.parse_variant_cast::<Bool>(Precedence::default());
-
-            parser.parse_variant("Int", Precedence::default(), |p| {
-                let n: u128 = p.number()?;
-                p.expect_char('_')?;
-                let ty: Ty = p.nonterminal()?;
-                Ok(ConstData::Value(Scalar::new(n).upcast(), ty))
-            });
```
nikomatsakis, line 135: I guess we could do the custom parse like this-- it's a bit odd because it will *also* accept `0 _ u8` but that's ok.

PR #241, opened 2026-02-10
Merging mergers 2

BoxyUwU:
WIP?


crates/formality-rust/src/grammar/binder.rs
```diff
@@ -258,15 +258,44 @@ where
...
-            write!(f, "{:?}", kind)?;
+            write!(f, "> ")?;
         }
-        write!(f, "> ")?;
         write!(f, "{:?}", &self.term)?;
         Ok(())
     }
 }
+
+impl<T> CoreParse<FormalityLang> for Binder<T>
```
nikomatsakis: wait I meant to dig into why this is here

nikomatsakis: oh right this file is dead

Issue #226, opened 2026-01-14
Handle `ensures` in `struct AliasBoundDeclBoundData` correctly


tiif:
It is currently encoded as something like `<T> [T: Foo]` where `T` represents the alias.

PR #242, opened 2026-02-11
refactor: make Decls compute proof data on-the-fly from Program grammar

nikomatsakis:
Replace pre-lowered declaration fields in Decls with on-the-fly computation from the Program grammar. Decls now contains only `program: Program` and `max_size: usize`. All intermediate lowering code in prove.rs is removed; lookup methods (trait_decl, impl_decls, neg_impl_decls, alias_eq_decls, adt_decl, trait_invariants, is_local_*) now live directly on Decls and walk the Program. Unit tests updated to construct grammar-level items (Trait, TraitImpl, etc.) instead of proof-level declarations.

This still isn't as beautiful as I might like it to be, but it's a net win I think.

### DIFF
--- a/crates/formality-rust/src/prove/prove/decls.rs
+++ b/crates/formality-rust/src/prove/prove/decls.rs
@@ -12,7 +12,6 @@ use formality_macros::term;
 pub struct Decls {
     pub program: Program,
     pub max_size: usize,
-
 }
 
 impl Decls {
@@ -120,7 +119,13 @@ impl Decls {
     }
 
     fn grammar_trait_to_decl(t: &Trait) -> TraitDecl {
-        let (vars, TraitBoundData { where_clauses, trait_items: _ }) = t.binder.open();
+        let (
+            vars,
+            TraitBoundData {
+                where_clauses,
+                trait_items: _,
+            },
+        ) = t.binder.open();
         TraitDecl {
             safety: t.safety.clone(),
             id: t.id.clone(),
--- a/crates/formality-rust/src/prove.rs
+++ b/crates/formality-rust/src/prove.rs
@@ -1,6 +1,4 @@
-use crate::grammar::{
-    Program, WhereBound, WhereBoundData, WhereClause, WhereClauseData,
-};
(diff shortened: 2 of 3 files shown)


crates/formality-rust/src/prove/prove/decls.rs
```diff
@@ -1,106 +1,272 @@
...
     }
 
-    pub fn trait_decl(&self, trait_id: &TraitId) -> &TraitDecl {
-        let mut v: Vec<_> = self
-            .trait_decls
-            .iter()
-            .filter(|t| t.id == *trait_id)
-            .collect();
-        assert!(!v.is_empty(), "no traits named `{trait_id:?}`");
-        assert!(v.len() <= 1, "multiple traits named `{trait_id:?}`");
```
tiif, line 59: We no longer have nice message for

```
        assert!(!v.is_empty(), "no traits named `{trait_id:?}`");
        assert!(v.len() <= 1, "multiple traits named `{trait_id:?}`");
```
?

We can just simply add this back in, or find a better place for it.

crates/formality-rust/src/prove/prove/decls.rs
```diff
@@ -260,30 +426,6 @@ pub struct AliasEqDeclBoundData {
...
-    pub fn alias_name(&self) -> AliasName {
-        self.binder.peek().alias.name.clone()
-    }
-}
-
-#[term($alias : $ensures $:where $where_clause)]
-pub struct AliasBoundDeclBoundData {
-    pub alias: AliasTy,
-    // FIXME(#226): This is currently encoded as something like `<T> [T: Foo]` where
-    // `T` represents the alias.
```
tiif, line 282: reminder to whoever see this later (including me), close #226 after this is merged.

crates/formality-rust/src/prove/prove/decls.rs
tiif: Actually it's fine,  ``trait_named`` will do the check :)

crates/formality-rust/src/prove/prove/decls.rs
lqd: Done!

PR #244, opened 2026-02-25
remove last vestiges of the old check

nikomatsakis:


src/test/borrowck.rs
```diff
@@ -159,51 +153,22 @@ fn min_problem_case_3() {
...
-                  }, Parameter::Const(_) => panic!("cannot outlive a constant"),
-              })`
-
-            the rule "write-indirect" at (nll.rs) failed because
-              pattern `TypedPlaceExpressionKind::Deref(place_loaned_ref)` did not match value `local(m)`
-
-            the rule "write-indirect" at (nll.rs) failed because
-              condition evaluted to false: `place_accessed.is_prefix_of(&place_loaned_ref.to_place_expression())`
-                place_accessed = *(local(m))
-                &place_loaned_ref.to_place_expression() = local(m)"#]]
```
tiif, line 206: It's interesting that this problem case 3 test passed. I suppose it is related to the change where we move ``verify_universal_outlives`` to ``prove_judgment``, but I don't really understand why.

crates/formality-rust/src/check/mini_rust_check.rs
tiif: We should be able to remove this function, it is no longer used.

tiif (commented):
I like the new function argument syntax :)

The final commit message is different from the diff, did you forget to add the change?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/244/72eac04ecf4adf1f78fda7af2b49ff87056a49be..f3f57e9a32f9a261055ebcb495c3cc2785d54e24)*

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -134,19 +135,9 @@ enum AccessKind {
 /// Given the `TypeckEnv`, and a (populated) list of `pending_outlives`
```
lqd: we can also remove this mention of `pending_outlives` since it's removed in this commit

src/test/mir_typeck.rs
```diff
@@ -509,13 +509,13 @@ fn test_incompatible_return_type() {
...
 
-        expect_test::expect!["judgment had no applicable rules: `prove { goal: {() <: u32}, assumptions: {}, env: Env { variables: [], bias: Soundness, pending: [], allow_pending_outlives: false }, decls: decls([crate Foo { fn foo (()) -> u32 = minirust(v1) -> v0 { let v0 : () ; let v1 : () ; exists { bb0 : { statements{ local(v0) = load(local(v1)) ; } return ; } } } ; }], 222) }`"]
+        expect_test::expect!["judgment had no applicable rules: `prove { goal: {() <: u32}, assumptions: {}, env: Env { variables: [], bias: Soundness, pending: [], allow_pending_outlives: false }, decls: decls([crate Foo { fn foo (v1 : ()) -> u32 = minirust(v1) -> v0 { let v0 : () ; let v1 : () ; exists { bb0 : { statements{ local(v0) = load(local(v1)) ; } return ; } } } ; }], 222) }`"]
     )
 }
 
 #[test]
 fn test_function_arg_number_mismatch() {
-    crate::assert_err!(
+    crate::assert_ok!(
```
lqd: shouldn't this still fail?

crates/formality-rust/src/grammar/minirust.rs
```diff
@@ -81,19 +70,11 @@ id!(FieldId);
 /// * For all `'a`
 ///     * There exists lifetimes `r0`, `r1`
 ///         * Such that the body is well-typed
-#[term(minirust($,args) -> $ret {
-    // First declare types etc parameters and return values
-    $*params
-
+#[term(minirust {
     // then declare the body
```
lqd: "the body" instead?

src/test/borrowck.rs
lqd: same, I don't understand why this changed either, nor why checking for universal subset errors more or at a different place would change anything here.

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -18,7 +17,7 @@ use crate::check::{
         places_live_before_basic_blocks, places_live_before_terminator, Assignment, LiveBefore,
         LivePlaces,
     },
-    mini_rust_check::{ty_is_adt, ty_is_ref, Location, PendingOutlives, TypeckEnv},
+    mini_rust_check::{ty_is_adt, ty_is_int, ty_is_ref, Location, PendingOutlives, TypeckEnv},
```
lqd: did `retcon` fail in this commit? the message looks more about the previous commit than this one

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -17,7 +18,7 @@ use crate::check::{
         places_live_before_basic_blocks, places_live_before_terminator, Assignment, LiveBefore,
         LivePlaces,
     },
-    mini_rust_check::{ty_is_adt, ty_is_int, ty_is_ref, Location, PendingOutlives, TypeckEnv},
+    mini_rust_check::{ty_is_adt, ty_is_ref, Location, PendingOutlives, TypeckEnv},
```
lqd, line 21: this is the inverse of the previous commit? "fix: check switch discriminant is an integer type in NLL"? :/

lqd (commented):
- +1 on tiif's comments
- the new syntax is way better
- the last 2 commits also likely need to be dropped
- It'd be cool to discuss the change we don't understand at the next meeting

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/244/72eac04ecf4adf1f78fda7af2b49ff87056a49be..f3f57e9a32f9a261055ebcb495c3cc2785d54e24)*

src/test/mir_typeck.rs
nikomatsakis: this test gets removed entirely, it's testing for a mismatch that just can't structurally occur anymore

src/test/borrowck.rs
nikomatsakis: The reason that problem case number 3 passes now is because we are no longer collecting all the outlives relations in a pre-pass and then using them everywhere along the graph, but rather accumulating them as we go. This is the heart of polonius, basically, and so it fixes this behavior.

lqd:
I was surprised this wasn't merged yet, especially because I mistakenly said it was in our call yesterday, apologies.

So I've made CI green and took care of 

> but to open a FIXME about adding back in the outlives constraints needed to accurately model NLL-as-it-is-today.

Done in https://github.com/rust-lang/a-mir-formality/issues/253

PR #245, opened 2026-03-03
parse all possibilities — propagate ambiguity through the parser

nikomatsakis:
## Summary

Three changes to formality-core's parser, developed together:

### (A) Parse all possibilities

Previously, the parser disambiguated eagerly at every nonterminal boundary — `Parser::finish()` compared reduction sequences via `is_preferable()` to pick the single "best" parse, or panicked on ambiguity. This meant ambiguity in a child nonterminal was invisible to its parent, preventing grammars where disambiguation needs context from higher up the parse tree.

Now the parser returns all successful parses and propagates them upward. Disambiguation happens once, at the top level (`core_term_with`), where the full picture is available. The implementation uses continuation-passing: `p.each_nonterminal(|x, p| { ... })` replaces the old `let x = p.nonterminal()?`. Each successful child parse forks execution for the parent — two ambiguous children produce a cross-product of results. The entire `ReductionKind` / `mark_as_identifier` / `is_preferable` disambiguation machinery is removed.

### (B) Scoped variable rejection for `id!` types

Previously, every `id!` type rejected identifiers matching a variable in scope. This is too aggressive — in a grammar like `foo::X`, the path segment after `::` shouldn't care that `X` happens to be a variable name. With parse-all-possibilities, the right answer falls out naturally: the `#[variable]` variant is preferred eagerly when it matches, and `id!` types just parse identifiers without checking scope. For the cases that *do* need to reject variables (to avoid genuine ambiguity), `id!(Name, match_var = false)` opts in explicitly.

### (C) Customizable binding syntax via `CoreParseBinding`

Previously, bindings were always parsed as `$kind $ident` (e.g., `ty T`, `lt a`). A new `CoreParseBinding<L>` trait on `L::Parameter` lets languages customize both binding parsing and variable reference syntax. formality-rust uses this to support `<T, 'a>` — the tick prefix infers lifetime kind, bare identifiers infer type kind, and `const` prefix infers const kind. No mandatory kind prefixes. All Rust-specific parsing stays in formality-rust; formality-core remains generic.

## Test plan

- All existing tests pass with updated syntax
- New `ambiguity.rs` tests verify deep ambiguity propagation and cross-product behavior
- New `ambiguity_helpers.rs` adds 27 focused tests covering every `each_*` combinator with ambiguous nonterminals
- `parser_var_id_ambiguity.rs` tests the `match_var = false` mechanism

### DIFF
--- a/crates/formality-macros/src/parse.rs
+++ b/crates/formality-macros/src/parse.rs
@@ -12,8 +12,29 @@ use crate::{
     variable::Variable,
 };
 
-/// Derive the `Parse` impl, using an optional grammar supplied "from the outside".
-/// This is used by the `#[term(G)]` macro, which supplies the grammar `G`.
+/// Derives a `CoreParse` impl for a struct or enum.
+///
+/// For each variant, generates code that tries to parse the input text and
+/// returns all successful parses (ambiguity is propagated upward, not resolved here).
+///
+/// `#[variable]` variants are tried first and short-circuit on success, ensuring
+/// in-scope variables are preferred over identifiers. For example, given:
+///
+/// ```rust,ignore
+/// #[term]
+/// pub enum TyData {
+///     #[cast]
+///     RigidTy(RigidTy),
+///     #[variable(ParameterKind::Ty)]
+///     Variable(Variable),
+/// }
+/// ```
+///
+/// If `T` is a type variable in scope, it parses as `Variable(T)` without
+/// trying `RigidTy`.
+///
+/// The `external_spec` parameter is used by `#[term(grammar)]` on structs
+/// to supply the grammar from outside rather than from a `#[grammar]` attribute.
 pub(crate) fn derive_parse_with_spec(
     mut s: synstructure::Structure,
     external_spec: Option<&FormalitySpec>,
@@ -87,6 +108,33 @@ pub(crate) fn derive_parse_with_spec(
     }))
 }
 
+/// Generates the parsing code for a single variant, choosing a strategy based on attributes.
+///
+/// There are four cases:
+///
+/// 1. **`#[grammar(...)]` or external spec** — parse according to the spec. See
+///    [`parse_variant_with_attr`] for details.
+///
+/// 2. **`#[variable(kind)]`** — parse a variable reference from scope:
+///    ```rust,ignore
+///    #[variable(ParameterKind::Ty)]
+///    Variable(Variable),
+///    // input "T" (where T is in scope) => Variable(T)
+///    ```
+///
+/// 3. **`#[cast]`** — transparent parse-through, no keyword prefix:
+///    ```rust,ignore
+///    #[cast]
+///    RigidTy(RigidTy),
+///    // input "u32" => RigidTy(RigidTy { name: u32, .. })
+///    ```
+///
+/// 4. **Default** (no attributes, has fields) — parse as `snake_case_name(field0, field1, ...)`:
+///    ```rust,ignore
+///    FnCall(Expr, Vec<Expr>),
+///    // input "fn_call(x, y)" => FnCall(x, y)
+///    ```
+///    Fieldless variants parse as just the keyword: `Foo` parses `"foo"`.
 fn parse_variant(
     variant: &synstructure::VariantInfo,
     external_spec: Option<&FormalitySpec>,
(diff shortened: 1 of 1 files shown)


nikomatsakis:
One thing -- this PR includes the code that finds a "best choice" amongst ambiguous choices, but I'm not sure if that's still needed. It doesn't seem bad but I imagine almost always that ambiguity is resolved in other ways. So probably it could be ripped out -- and maybe we don't even need to track reduction trees anymore? I will think about that.

crates/formality-core/src/parse/parser/left_recursion.rs
```diff
@@ -383,59 +383,61 @@ where
...
         tracing::trace!(
-            "reparsing of left-recursive grammar: values = {:#?}",
-            values
+            "reparsing of left-recursive grammar: all_values = {:#?}",
+            all_values
         );
 
-        // If we have an intermediate value, update the stack entry to point at.
-        // This takes a borrow of `value` but converts it into a raw pointer.
+        // Pick the "best" (longest consumed) parse as the seed for left-recursion reuse.
```
nikomatsakis, line 410: Hmm, re-reading this, I'm not 100% sure this is the right thing. It probably is in practice. I should try to make some kind of torture test though :)

crates/formality-core/src/parse.rs
```diff
@@ -285,7 +273,9 @@ where
 {
     fn parse<'t>(scope: &Scope<L>, text: &'t str) -> ParseResult<'t, Self> {
         Parser::single_variant(scope, text, "Set", |p| {
-            p.delimited_nonterminal('{', false, '}')
+            p.each_delimited_nonterminal('{', false, '}', |v: Vec<T>, p| {
```
nikomatsakis, line 280: this is an example of a typical change; we don't just parse once, we need closures to account for ambig possibilities

crates/formality-core/src/parse.rs
```diff
@@ -489,20 +511,26 @@ impl<L: Language, A: CoreParse<L>, B: CoreParse<L>, C: CoreParse<L>> CoreParse<L
...
-            p.expect_char(',')?;
-            let c: C = p.nonterminal()?;
-            p.skip_trailing_comma();
-            p.expect_char(')')?;
-            Ok((a, b, c))
+            p.each_nonterminal(|a: A, p| {
+                p.expect_char(',')?;
+                p.each_nonterminal(|b: B, p| {
+                    p.expect_char(',')?;
+                    p.each_nonterminal(|c: C, p| {
```
nikomatsakis, line 522: another example

crates/formality-macros/src/parse.rs
```diff
@@ -40,16 +40,37 @@ pub(crate) fn derive_parse_with_spec(
         }
     }
 
-    let mut parse_variants = TokenStream::new();
+    // Emit `#[variable]` variants first. If any succeeds, short-circuit
```
nikomatsakis, line 64: this is where the variable precedence is implemented

crates/formality-rust/src/prove/prove/test/adt_wf.rs
```diff
@@ -10,7 +10,7 @@ fn decls() -> Decls {
         program: Decls::program_from_items(vec![
             term("trait Foo where {}"),
             term("impl Foo for u32 {}"),
-            term("struct X<ty T> where T : Foo {}"),
+            term("struct X<T> where T : Foo {}"),
```
nikomatsakis, line 13: an example of more Rust-like syntax

src/test/borrowck.rs
```diff
@@ -141,18 +141,18 @@ fn min_problem_case_3() {
...
                 = minirust(m) -> ret {
-                    let ret: &mut a Map;
-                    let m: &mut a Map;
+                    let ret: &mut 'a Map;
+                    let m: &mut 'a Map;
 
-                    exists<lt r0, lt r1> {
-                        let n: &mut r0 Map;
-                        let o: &mut r1 Map;
+                    exists<'r0, 'r1> {
```
nikomatsakis, line 149: lifetimes!

src/test/borrowck.rs
```diff
@@ -334,13 +334,13 @@ fn declared_transitive_universal_region_relationship() {
...
                 where
-                    a: b,
-                    b: c,
+                    'a: 'b,
+                    'b: 'c,
                 = minirust(v1) -> v0 {
-                    let v0: &c u32;
-                    let v1: &a u32;
+                    let v0: &'c u32;
+                    let v1: &'a u32;
```
nikomatsakis, line 343: more lifetimes!

src/test/borrowck.rs
```diff
@@ -531,36 +531,36 @@ fn cfg_union_approx_cause_false_error() {
...
+                                local(p) = &mut 'loan_1 local(a);
+                                local(q) = &mut 'loan_2 local(b);
                             }
                             goto bb3;
                         }
 
                         bb2: {
                             statements {
-                                local(p) = &mut loan_3 local(b);
+                                local(p) = &mut 'loan_3 local(b);
```
nikomatsakis, line 563: (this I didn't fix yet)

tests/parser-torture-tests/ambiguity.rs
```diff
@@ -52,16 +52,105 @@ fn reduce_reduce_ambig() {
...
+    // We should get exactly 2 complete parses: one via InnerA, one via InnerB.
+    assert_eq!(
+        complete.len(),
+        2,
+        "expected 2 ambiguous parses, got {}: {complete:#?}",
+        complete.len()
+    );
+}
+
+/// Test that deep ambiguity through a sequence of two nonterminals
```
nikomatsakis, line 117: some focused tests

crates/formality-core/src/parse/parser/left_recursion.rs
nikomatsakis: Yeah, this is bogus. I pushed a commit showing the problem but left it as a FIXME.

nikomatsakis: See #251

crates/formality-core/src/parse/parser.rs
```diff
@@ -232,7 +232,7 @@ where
         })
     }
 
-    /// Like [`Self::identifier_nonterminal`], but rejects identifiers that match a variable in scope.
+    /// Like [`Self::identifier`], but rejects identifiers that match a variable in scope.
```
nikomatsakis, line 235: (this commit seemed like fixup but it couldn't be easily rebased and I kinda gave up)

crates/formality-core/src/parse.rs
```diff
@@ -84,13 +85,28 @@ where
...
+        .into_iter()
+        .filter(|p| skip_whitespace(p.text()).is_empty())
+        .collect();
+
+    if complete_parses.is_empty() {
+        crate::bail!("extra tokens after parsing {text:?}");
+    }
+
+    // Deduplicate parses that produced the same value.
+    complete_parses.dedup_by(|a, b| a.value == b.value);
```
lqd: It's not a big deal since it'll just reduce the number of parsing ambiguities (AFAICT at this point of my review, I don't know what future this PR holds for me 🔮), but wanted to note this will only deduplicate consecutive duplicate parses.

crates/formality-macros/src/parse.rs
```diff
@@ -173,73 +198,194 @@ fn parse_variant_with_attr(
...
-            spec::FormalitySpecSymbol::Char { punct } => {
-                let literal = Literal::character(punct.as_char());
-                quote_spanned!(punct.span() =>
-                    __p.expect_char(#literal)?;
-                )
-            }
+/// Recursively wrap a list of spec symbols, nesting `each_nonterminal` for `FieldMode::Single`
+/// fields and keeping other operations flat.
+///
+/// `captured_names` tracks variable names bound by outer `each_nonterminal` calls.
```
lqd: what `captured_names`? I guess this partially describes `wrap_symbols_inner` as well. 

But also `wrap_symbols_in_each_nonterminal` is used only once, it's not really needed :3

crates/formality-core/src/parse.rs
nikomatsakis: Yeah, this .. seems kind of pointless doesn't it?

PR #246, opened 2026-03-05
All refs all the fricking time

nikomatsakis:
This branch cleans up the judgment rules a lot. The basic idea is that we introduce `&` borrows into key places and leverage the match ergonomics rules so that all the variables are `&`-references. This makes the borrow checker a non-issue. The result is that things read much more nicely.

### DIFF
--- a/crates/formality-rust/src/prove/prove/prove/negation.rs
+++ b/crates/formality-rust/src/prove/prove/prove/negation.rs
@@ -106,11 +106,7 @@ pub fn negation_via_failure<T: CoreFold<FormalityLang>>(
                 let label = format!("{:?}", result);
                 ProvenSet::singleton((
                     result,
-                    ProofTree::new(
-                        label,
-                        Some("ambiguous_negation"),
-                        vec![sample_tree.clone()],
-                    ),
+                    ProofTree::new(label, Some("ambiguous_negation"), vec![sample_tree.clone()]),
                 ))
             }
         }
--- a/crates/formality-rust/src/prove/prove/prove/prove_outlives.rs
+++ b/crates/formality-rust/src/prove/prove/prove/prove_outlives.rs
@@ -83,7 +83,10 @@ judgment_fn! {
 
 /// Given a region `r1`, find a set of all regions `r2` where `r1 : r2` transitively
 /// according to the assumptions.
-fn transitively_outlived_by(assumptions: impl Upcast<Wcs>, r1: impl Upcast<Parameter>) -> Set<Parameter> {
+fn transitively_outlived_by(
+    assumptions: impl Upcast<Wcs>,
+    r1: impl Upcast<Parameter>,
+) -> Set<Parameter> {
     let assumptions: Wcs = assumptions.upcast();
     let r1: Parameter = r1.upcast();
     let mut reachable = Set::new();
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -195,14 +195,9 @@ impl Check<'_> {
         assert!(env.only_universal_variables());
         assert!(env.encloses((&assumptions, &goal)));
 
-        let cs = is_definitely_not_proveable(
-            env,
-            &assumptions,
-            &goal,
-            |env, assumptions, goal| {
-                crate::prove::prove::prove(self.decls, env, &assumptions, &goal)
-            },
-        );
+        let cs = is_definitely_not_proveable(env, &assumptions, &goal, |env, assumptions, goal| {
+            crate::prove::prove::prove(self.decls, env, &assumptions, &goal)
+        });
         let cs = cs.into_map()?;
         if let Some((_, proof_tree)) = cs.iter().find(|(c, _)| c.unconditionally_true()) {
             return Ok(proof_tree.clone());


crates/formality-core/src/judgment.rs
```diff
@@ -83,20 +83,27 @@ macro_rules! judgment_fn {
...
+                )*
+                $(
+                    $crate::respan!(
+                        $assert_expr
+                        (
+                            formality_core::judgment::JudgmentAssertion::assert($assert_expr, stringify!($assert_expr));
+                        )
+                    );
+                )*
+            }
```
nikomatsakis, line 106: Reworked `(assert)` sections here because they were given "raw" access to the parameters of the function, which are owned types.

Reading over it now, it might be good to move the `&` up into the definition of the local variables though.

crates/formality-core/src/judgment.rs
```diff
@@ -313,21 +320,21 @@ macro_rules! push_rules {
 
     (@match $conclusion_name:ident inputs($in0:ident $($inputs:tt)*) patterns($pat0:ident : $ty0:ty, $($pats:tt)*) args $args:tt) => {
         {
-            if let Some($pat0) = $crate::Downcast::downcast::<$ty0>($in0) {
+            if let Some($pat0) = &$crate::Downcast::downcast::<$ty0>($in0) {
```
nikomatsakis, line 323: Most of the macro changes were like this -- introduce an `&` and lean on match ergonomics to propagate that into any bindings.

crates/formality-core/src/judgment.rs
```diff
@@ -313,21 +320,21 @@ macro_rules! push_rules {
...
                 $crate::push_rules!(@match $conclusion_name inputs($($inputs)*) patterns($($pats)*) args $args);
             }
         }
     };
 
     (@match $conclusion_name:ident inputs($in0:ident $($inputs:tt)*) patterns($pat0:ident, $($pats:tt)*) args $args:tt) => {
         {
-            let $pat0 = Clone::clone($in0);
+            let $pat0 = $in0;
```
nikomatsakis, line 331: Here, `$in0` was already a reference, so we just skip cloning. Same pattern.

crates/formality-core/src/judgment.rs
```diff
@@ -313,21 +320,21 @@ macro_rules! push_rules {
...
             $crate::push_rules!(@match $conclusion_name inputs($($inputs)*) patterns($($pats)*) args $args);
         }
     };
 
     (@match $conclusion_name:ident inputs($in0:ident $($inputs:tt)*) patterns($pat0:pat, $($pats:tt)*) args $args:tt) => {
-        if let Some($pat0) = $crate::Downcast::downcast(&$in0) {
+        if let Some($pat0) = &$crate::Downcast::downcast($in0) {
```
nikomatsakis, line 337: Introduce a reference to the result of downcasting, same pattern.

crates/formality-core/src/judgment.rs
```diff
@@ -357,7 +364,7 @@ macro_rules! push_rules {
     ) => {
         match $crate::judgment::try_catch(|| Ok($e)) {
             Ok(value) => {
-                if let $p = Clone::clone(&value) {
+                if let $p = &value {
```
nikomatsakis, line 367: Same pattern: don't clone.

crates/formality-core/src/judgment.rs
```diff
@@ -478,15 +485,15 @@ macro_rules! push_rules {
         ($($argn:ident = $arge:expr),*); $cond:expr; $origcond:expr; $($m:tt)*
     ) => {
         $(
-            let $argn = &$arge;
+            let $argn = $arge;
```
nikomatsakis, line 488: In this case, the structured-if values were accumulated as references.

crates/formality-core/src/judgment.rs
```diff
@@ -625,7 +632,9 @@ macro_rules! push_rules {
     ) => {
         if let Err(e) = $crate::judgment::EachProof::each_proof(
             $i,
-            |($p, proof_tree)| {
+            |(value, proof_tree)| {
+                let $p = &value;
```
nikomatsakis, line 636: Again, apply `&` and lean on match-ergonomics (and hygiene...).

crates/formality-core/src/judgment.rs
```diff
@@ -654,7 +663,7 @@ macro_rules! push_rules {
         match $crate::judgment::try_catch::<$t>(|| Ok($i)) {
             Ok(p) => {
                 let proof_tree = $crate::judgment::ProofTree::leaf(format!("{} = {p:?}", stringify!($p)));
-                let $p = p;
+                let $p = &p;
```
nikomatsakis, line 666: Same pattern.

crates/formality-core/src/judgment.rs
```diff
@@ -672,7 +681,7 @@ macro_rules! push_rules {
         match $crate::judgment::try_catch(|| Ok($i)) {
             Ok(p) => {
                 let proof_tree = $crate::judgment::ProofTree::leaf(format!("{} = {p:?}", stringify!($p)));
-                let $p = p;
+                let $p = &p;
```
nikomatsakis, line 684: Same pattern.

crates/formality-core/src/judgment.rs
```diff
@@ -692,7 +701,9 @@ macro_rules! push_rules {
         if let Err(e) = $crate::judgment::member_of(
             $i,
             || stringify!($i).to_string(),
-            |($p, proof_tree)| {
+            |(value, proof_tree)| {
+                let $p = &value;
```
nikomatsakis, line 705: Same pattern.

crates/formality-rust/src/prove/prove/prove/negation.rs
```diff
@@ -57,7 +57,7 @@ pub fn may_not_be_provable<T: CoreFold<FormalityLang>>(
 pub fn negation_via_failure<T: CoreFold<FormalityLang>>(
     env: &Env,
     assumptions: impl Upcast<Wcs>,
-    data: T,
+    data: &T,
```
nikomatsakis, line 60: In a few places, take `&` on functions instead of ownership. Sometimes also `impl Upcast`, but `&T` is better for type inference.

crates/formality-rust/src/prove/prove/prove/prove_const_has_type.rs
```diff
@@ -29,9 +29,9 @@ judgment_fn! {
         (
             (let minirust::Body { ret, params, .. } = body)
             (p in params)
-            (if p.id == ret)
+            (if p.id == *ret)
```
nikomatsakis, line 32: Gotta love Rust's persnickety behavior around `==` and `&`.

crates/formality-rust/src/prove/prove/prove/prove_eq.rs
```diff
@@ -145,12 +145,16 @@ judgment_fn! {
...
 
 fn equate_variable(
-    decls: Decls,
-    mut env: Env,
-    assumptions: Wcs,
-    x: ExistentialVar,
+    decls: impl Upcast<Decls>,
+    env: impl Upcast<Env>,
+    assumptions: impl Upcast<Wcs>,
+    x: impl Upcast<ExistentialVar>,
```
nikomatsakis, line 151: Here's an example of `impl Upcast`

crates/formality-rust/src/prove/prove/prove/prove_outlives.rs
```diff
@@ -83,7 +83,12 @@ judgment_fn! {
 
 /// Given a region `r1`, find a set of all regions `r2` where `r1 : r2` transitively
 /// according to the assumptions.
-fn transitively_outlived_by(assumptions: Wcs, r1: Parameter) -> Set<Parameter> {
+fn transitively_outlived_by(
+    assumptions: impl Upcast<Wcs>,
+    r1: impl Upcast<Parameter>,
```
nikomatsakis, line 88: Another `impl Upcast` addition.

tests/judgment-error-reporting/cyclic_judgment.rs
```diff
@@ -33,7 +33,7 @@ judgment_fn! {
         )
 
         (
-            (sub(&*a, &*b) => ())
+            (sub(Ty::clone(a), Ty::clone(b)) => ())
```
nikomatsakis, line 36: (Hmm, this could probably just be `a` and `b`! I should fix this.)

tests/judgment-error-reporting/cyclic_judgment.rs
```diff
@@ -51,13 +51,13 @@ judgment_fn! {
 
         (
             ----------------------------- ("add my")
-            (equivalent(Ty::Class { name }) => Ty::My(Arc::new(Ty::Class { name })))
+            (equivalent(Ty::Class { name }) => Ty::My(Arc::new(Ty::Class { name: name.clone() })))
```
nikomatsakis, line 54: (Similarly, this could maybe be `Ty::class()`? Gotta fix this.)

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -307,28 +307,28 @@ judgment_fn! {
...
+            (let this_entry = StackEntry::new(env, fn_assumptions, loans_live_on_entry, outlives, bb_id))
             (if !stack.contains(&this_entry))!
-            (let stack = { let mut s = stack.clone(); s.push(this_entry); s })
-            (let BasicBlock { id: _, statements, terminator } = env.basic_block(&bb_id)?)
-            (terminator_wf(&env, &terminator) => ())
-            (let places_live_before_terminator = places_live_before_terminator(&env, &terminator))
+            (let stack = { let mut s = stack.clone(); s.push(this_entry.clone()); s })
+            (let BasicBlock { id: _, statements, terminator } = env.basic_block(bb_id)?)
+            (terminator_wf(env, terminator) => ())
+            (let places_live_before_terminator = places_live_before_terminator(env, terminator))
```
nikomatsakis, line 322: (I want to clean this up, but I drew the line here in terms of making more changes to make it more readable.)

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -789,16 +787,16 @@ judgment_fn! {
             --- ("local")
             (borrow_check_place_expression(env, _assumptions, loans_live, outlives, PlaceExpression::Local(local_id), _places_live) => (
                 TypedPlaceExpression {
-                    ty,
-                    kind: TypedPlaceExpressionKind::Local(local_id),
+                    ty: ty.clone(),
```
nikomatsakis, line 790: (Unfortunate, but another thing I planned to revisit in the future, we'd need an `impl Upcast` helper.)

crates/formality-rust/src/prove/prove/prove/prove_after.rs
```diff
@@ -15,7 +15,8 @@ judgment_fn! {
         debug(constraints, goal, assumptions)
 
         (
-            (let (assumptions, goal) = c1.substitution().apply(&(Wcs::clone(assumptions), Wcs::clone(goal))))
+            (let assumptions = c1.substitution().apply(assumptions))
+            (let goal = c1.substitution().apply(goal))
```
lqd, line 19: I guess `(let (assumptions, goal) = c1.substitution().apply(&(assumptions, goal)))` doesn't work?

nikomatsakis: Actually, I think that does work, or could be made to work.

nikomatsakis: Let me look a bit closer.

nikomatsakis: Oh, right, I remember now. This doesn't work because `CoreFold` follows the rustc pattern where the output type is `Self`. I wanted to change that but decided it was out-of-scope for this PR.

lqd: Ok 👍

crates/formality-core/src/judgment.rs
nikomatsakis: Did a bit of investigation and this is trickier than it appears, the owned values are moved into a struct later on.

PR #252, opened 2026-03-11
Add tests

tiif:
Add some tests for the "lifetime" rule in ``loan_not_required_by_parameter``.

### DIFF
--- a/src/test/borrowck.rs
+++ b/src/test/borrowck.rs
@@ -552,3 +552,120 @@ fn cfg_union_approx_cause_false_error() {
         ]
     )
 }
+
+/// Fail test for loan_cannot_outlive's "lifetime" rule.
+/// This is equivalent to:
+/// ```
+/// fn foo() -> u32 {
+///     let mut x: u32 = 22;
+///     let p = &x;
+///     let q = p;
+///     x += 1;
+///     println!("{q}");
+///     return 0;
+/// }
+/// ```
+#[formality_core::test]
+fn loan_cannot_outlive_lifetime_fail() {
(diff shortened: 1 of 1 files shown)


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -1215,11 +1215,11 @@ judgment_fn! {
...
             //
             // ```
-            // let mut x = 22;
+            // let mut x: u32 = 22;
             // let p = &'0 x;
-            // let q: &'1 x = p; // requires `'0: '1`
+            // let q: &'1 u32 = p; // requires `'0: '1`
             // x += 1;
-            // use(q); // could this use data derived from `p`? (obviously yes)
+            // use(q); // could this use data derived from `p`? (obviously no)
```
lqd: Could you say more about this change? To me this is saying we have a use of q, which is itself derived from p (itself derived from x), so """q/p/x are live at the use, and thus the prior x+=1 is an error"""

nikomatsakis: I agree that the comment seems correct as written

tiif: Ok interesting... how I see it is, if we don't have ``use(q)`` after writing to ``x``, we won't have any error. We can keep the comment and just add the test.

lqd: Exactly, we won't have an error because neither p nor q would be live when writing to x. Which is what the comment was getting at, if there's something that could use the data, we can't allow it.

PR #256, opened 2026-03-19
rewrite to the new "expr-based" borrow checker

nikomatsakis:
This is a pretty major rewrite. I walked through it at a high-level with @lqd and @tiif. The long-and-short of it is that it rewrites from "minirust" to a simple stmt/expr language that can be *converted* into minirust (the actual minirust, from the minirust repo). It reimplements the borrow checker logic on this new form, fixing some bugs along the way. It also does some drive-by improvements like better parser diagnostics and so forth.

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/liveness.rs
+++ b/crates/formality-rust/src/check/borrow_check/liveness.rs
@@ -1,16 +1,134 @@
 use crate::{
     check::borrow_check::env::TypeckEnv,
-    grammar::expr::{Block, Expr, ExprData, FieldExpr, Init, PlaceExpr, Stmt},
+    check::borrow_check::flow_state::FlowState,
+    grammar::expr::{Block, Expr, ExprData, FieldExpr, Init, Label, LabelId, PlaceExpr, Stmt},
 };
 use formality_core::{Set, Upcast};
 
 pub type LivePlaces = Set<PlaceExpr>;
 
+/// Liveness context for control-flow targets (break/continue).
+/// Each entry corresponds to a labeled scope (block or loop).
+#[derive(Clone, Debug)]
+pub struct LivenessContext {
+    scopes: Vec<LivenessScope>,
+}
+
+#[derive(Clone, Debug)]
+struct LivenessScope {
+    label: Option<LabelId>,
+    /// Places live after this scope exits (break target).
+    break_live: LivePlaces,
+    /// Places live at loop entry (continue target). `None` for non-loop scopes.
+    continue_live: Option<LivePlaces>,
+}
+
+impl LivenessContext {
+    pub fn empty() -> Self {
+        LivenessContext { scopes: Vec::new() }
+    }
+
+    /// Look up the break target liveness for the given label.
+    fn break_live(&self, label: &LabelId) -> LivePlaces {
+        self.scopes
+            .iter()
+            .rev()
+            .find(|s| s.label.as_ref() == Some(label))
+            .map(|s| s.break_live.clone())
+            .unwrap_or_default()
+    }
+
+    /// Look up the continue target liveness for the given label.
+    fn continue_live(&self, label: &LabelId) -> LivePlaces {
+        self.scopes
+            .iter()
+            .rev()
+            .find(|s| s.label.as_ref() == Some(label))
+            .and_then(|s| s.continue_live.clone())
+            .unwrap_or_default()
+    }
+
+    /// Push a loop scope and return a new context.
+    fn with_loop(
+        &self,
+        label: &Option<Label>,
+        break_live: LivePlaces,
+        continue_live: LivePlaces,
+    ) -> Self {
+        let mut ctx = self.clone();
+        ctx.scopes.push(LivenessScope {
+            label: label.as_ref().map(|l| l.id.clone()),
+            break_live,
+            continue_live: Some(continue_live),
+        });
+        ctx
+    }
+
+    /// Update the continue liveness for the innermost scope.
+    fn set_continue_live(&mut self, live: LivePlaces) {
+        if let Some(scope) = self.scopes.last_mut() {
+            scope.continue_live = Some(live);
+        }
+    }
+
+    /// Push a block scope (no continue target).
+    fn with_block(&self, label: &Option<Label>, break_live: LivePlaces) -> Self {
+        let mut ctx = self.clone();
+        ctx.scopes.push(LivenessScope {
+            label: label.as_ref().map(|l| l.id.clone()),
+            break_live,
+            continue_live: None,
+        });
+        ctx
+    }
+}
+
+/// Trait for types that can be converted to a [`LivenessContext`].
+/// This allows `live_before` to accept either a `&FlowState` or a `&LivenessContext`.
+pub trait IntoLivenessContext {
+    fn into_liveness_context(&self) -> LivenessContext;
+}
+
+impl IntoLivenessContext for FlowState {
+    fn into_liveness_context(&self) -> LivenessContext {
+        LivenessContext {
+            scopes: self
+                .scopes
+                .iter()
+                .map(|s| LivenessScope {
+                    label: s.label.clone(),
+                    break_live: s.break_live_places.clone(),
+                    continue_live: s.continue_live_places.clone(),
+                })
+                .collect(),
+        }
+    }
+}
+
+impl<T: IntoLivenessContext> IntoLivenessContext for &T {
+    fn into_liveness_context(&self) -> LivenessContext {
+        T::into_liveness_context(self)
+    }
+}
+
+impl IntoLivenessContext for LivenessContext {
+    fn into_liveness_context(&self) -> LivenessContext {
+        self.clone()
+    }
+}
+
 pub trait LiveBefore {
     /// Given a term `term` and a set of places `places_live` that are live after `term` executes,
     /// returns the set of places live BEFORE `term`. This will include all places in `places_live`
     /// minus any places overwritten by `term`, plus any places accessed.
-    fn live_before(&self, env: &TypeckEnv, places_live: impl Upcast<LivePlaces>) -> LivePlaces;
+    ///
+    /// The `scopes` parameter provides liveness targets for break/continue statements.
+    fn live_before(
+        &self,
+        env: &TypeckEnv,
+        scopes: &impl IntoLivenessContext,
+        places_live: impl Upcast<LivePlaces>,
+    ) -> LivePlaces;
 }
 
 /// Sequential execution: `a` runs before `b`.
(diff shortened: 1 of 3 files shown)


crates/formality-core/src/parse.rs
```diff
@@ -170,22 +290,76 @@ where
...
     pub fn at(text: &'t str, message: String) -> Set<Self> {
         let text = parser::skip_whitespace(text);
-        set![ParseError { text, message }]
+
+        // Snapshot the nonterminal stack. Convert raw pointers back to &'t str.
+        // SAFETY: The start_text pointers in the stack are subslices of the
+        // original input &'t str, which is still alive — each StackEntry is
+        // only alive while its corresponding `enter` frame is active on the
+        // call stack, and we are inside such a frame.
+        let backtrace: Vec<ParseFrame<'t>> = parser::snapshot_nonterminal_stack()
```
nikomatsakis: Pre-existing but I'm not totally sure why I made this use thread-locals etc, it might be nice to refactor the parse APIs to take some context.

nikomatsakis: At minimum it would be good to move this unsafety into the `snapshot_nonterminal_stack` method so it's more self-contained.

crates/formality-core/src/parse/parser/left_recursion.rs
```diff
@@ -486,6 +492,34 @@ pub(crate) fn snapshot_nonterminal_stack() -> Vec<(&'static str, *const str)> {
...
 
+/// Record on the *current* stack frame that the text at `(scope, text)` was
+/// successfully probed as a variable.  Called from `Parser::try_claim_as_var`.
+pub(super) fn set_claimed_as_var<L: Language>(scope: &Scope<L>, text: &str) {
+    STACK.with_borrow_mut(|stack| {
+        if let Some(top) = stack.last_mut() {
+            let scope_ptr: *const () = erase_type(scope);
+            let text_ptr: *const str = text;
+            // Only set the flag when the pointers match (same parse position).
+            if scope_ptr == top.scope && std::ptr::eq(text_ptr, top.start_text) {
```
nikomatsakis: Hmm, I'm trying to remember, I think this could perhaps be an assert?

crates/formality-core/src/judgment/proven_set.rs
```diff
@@ -413,16 +422,45 @@ pub fn insert_smallest_proof<K: Ord + Clone>(
     }
 }
 
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone)]
```
nikomatsakis, line 426: Should probably add a doc comment to this. We are tracking the location in the source where the failure occurred.

crates/formality-core/src/judgment/proven_set.rs
```diff
@@ -453,9 +493,16 @@ impl FailedJudgment {
     /// returns `(set, has_non_cycle)`. The `set` represents a new, simplified set of failued
     #[tracing::instrument(level = "Debug", skip(stack), ret)]
     fn strip_cycles(
-        stack: Set<&String>,
+        stack: &Set<&String>,
         failed_rules: Set<FailedRule>,
     ) -> (Set<FailedRule>, HasNonCycle) {
+        // If we did not find any applicable rules at all,
+        // then this is not a cycle failure, it's just an
+        // unhandled case.
```
nikomatsakis: This comment is kinda unclear. I meant "It's a case where the input doesn't match any of the judgment rules."

crates/formality-core/src/judgment.rs
```diff
@@ -558,22 +562,22 @@ macro_rules! push_rules {
...
                 #[allow(unused)]
-                let ($($with_var,)*) = loop_carried.clone();
+                let ($($with_var,)*) = &loop_carried.clone();
 
                 let attributes = vec![("item".to_string(), format!("{:?}", $loop_var))];
                 let mut item_proof_trees = vec![];
-                let mut next_carried = None;
-                $crate::push_rules!(@body (loop(next_carried, ($($with_var,)*))); $inputs; item_proof_trees; $($inner_step)*);
+                let mut next_carried_and_proof = None;
+                $crate::push_rules!(@body (loop(next_carried_and_proof, ($($with_var,)*))); $inputs; item_proof_trees; $($inner_step)*);
```
nikomatsakis, line 573: Context for this: the proof tree with `for_all` was not actually *collecting* the proof trees for each iteration. That's what this change does, we propagate back the proof tree from the iteration together with the next values of any loop-carried values. Could be a good comment.

crates/formality-core/src/judgment.rs
```diff
@@ -633,21 +637,27 @@ macro_rules! push_rules {
         if let Err(e) = $crate::judgment::EachProof::each_proof(
             $i,
             |(value, proof_tree)| {
-                let $p = &value;
-
-                // Remember the size of the child proof tree stack
-                let len = $child_proof_trees.len();
+                #[allow(irrefutable_let_patterns)]
+                if let $p = &value {
```
nikomatsakis, line 644: Using `if let` here means you can do `(foo(...) => RefutablePattern { ... })` which is really nice

crates/formality-rust/src/prove/prove/test/eq_partial_eq.rs
```diff
@@ -30,7 +30,10 @@ fn not_partial_eq_implies_eq() {
     let goal: Wc = term("for<T> if {PartialEq(T)} Eq(T)");
     prove(decls(), (), (), goal)
     .assert_err(
-    expect!["judgment had no applicable rules: `prove { goal: {for <ty> if {PartialEq(^ty0_0)} Eq(^ty0_0)}, assumptions: {}, env: Env { variables: [], bias: Soundness, pending: [], allow_pending_outlives: false }, decls: decls([crate test { trait Eq <ty> where ^ty0_0 : PartialEq { } trait PartialEq <ty> { } }], 222) }`"]);
```
nikomatsakis, line 33: This change resulted from the changes to `strip_cycle`; previously this was giving a very imprecise error.

crates/formality-core/src/parse/parser.rs
```diff
@@ -827,31 +842,10 @@ where
...
-            "each_opt_nonterminal({}): also trying None branch",
-            std::any::type_name::<T>()
-        );
-        let mut forked = self.clone();
-        let none_result = op(None, &mut forked);
-        Self::combine_parse_results(some_result, none_result)
+        self.combine_parse_results(
+            op(None, &mut self.clone()),
+            self.each_nonterminal(|t: T, p| op(Some(t), p)),
+        )
```
nikomatsakis, line 848: These changes generalized the parser to truly try all possible parses. Needed to accept the full Expr grammar which would have more ambiguity otherwise.

crates/formality-macros/src/parse.rs
```diff
@@ -61,29 +61,24 @@ pub(crate) fn derive_parse_with_spec(
         }
     }
 
-    // Build all variant parsers in a single stream.
-    // If any variant carries `#[variable]`, we emit a `try_claim_as_var`
-    // probe *before* parsing any variants so that `id!` types will
-    // reject the same text as a plain identifier.
+    // Separate variable variants from non-variable variants.
+    // Variable variants are parsed first; if any succeed, we mark
+    // `claimed_as_var` on the left-recursion stack so that `id!` types
```
nikomatsakis: This comment is out of date. We don't parse all variable variants first anymore, rather, we invoke `try_claim_as_var` which will try to parse the string as a variable *of any kind*, but we only do that if there is at least one variable variant.

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub current: PointFlowState,
+
+    /// State to be incorporated at the end of the labeled block/loop/etc
+    pub breaks: Set<LabeledFlowState>,
+
+    /// State to be incorporated at a continue targeting the labeled block/loop/etc
+    pub continues: Set<LabeledFlowState>,
+}
+
+/*
```
nikomatsakis: Whoops, this was just me explaining stuff, we could remove it -- but maybe it should be turned into an mdbook chapter, actually.

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -0,0 +1,612 @@
...
+    /// Returns the locals that would be dropped when the innermost scope exits,
+    /// in reverse declaration order (LIFO).
+    pub fn locals_dropped_in_innermost_scope(&self) -> Vec<TypedPlaceExpr> {
+        self.scopes
+            .last()
+            .map(|scope| scope.drop_places.iter().rev().cloned().collect())
+            .unwrap_or_default()
+    }
+
+    pub fn pop_scope(&self, expected_label: &Option<Label>) -> Self {
```
nikomatsakis, line 407: Ought to have a doc-comment. This method removes the scope from the top of the stack and checks that its label is equal to `expected_label`. When we pop the stack, we also incorporate any control-flow that is pending in the `breaks` set for a block with this label.

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -0,0 +1,612 @@
...
+
+        FlowState {
+            scopes,
+            current: successor,
+            breaks,
+            continues: other_labels,
+        }
+    }
+
+    pub fn no_continues(&self, label: &Option<Label>) -> bool {
```
nikomatsakis: Ought to have a doc-comment Checks that there are no entries in the `continues` array with the given label. Also, this may be dead-code now, I vaguely recall removing the calls to it.

crates/formality-rust/src/check/borrow_check/liveness.rs
```diff
@@ -130,60 +72,118 @@ impl<T: LiveBefore> LiveBefore for Option<T> {
...
+                condition,
+                then_block,
+                else_block,
+            } => Seq(condition, Either(then_block, else_block)).live_before(env, places_live),
+            Stmt::Expr { expr } => expr.live_before(env, places_live),
+            Stmt::Loop { label: _, body } => {
+                // Fixed-point iteration: places live before the loop include
+                // places live after the loop plus any places the body needs.
+                let mut live = places_live;
+                loop {
```
nikomatsakis, line 275: Note to self: This liveness calculation seems insufficient. I thought we did better here. I think we have to account for incoming continue edges? This may get fixed in a later commit, or maybe I jsut *thought* about fixing it. Would be good to have an example.

crates/formality-rust/src/check/borrow_check/outlives.rs
```diff
@@ -0,0 +1,113 @@
...
+use crate::check::borrow_check::flow_state::PendingOutlives;
+
+use crate::grammar::{Parameter, Relation, Variable, Wcs};
+use crate::prove::prove::prove;
+use formality_core::{judgment_fn, Set, Upcast};
+
+judgment_fn! {
+    /// Verify that all pending outlives constraints between universal lifetime variables
+    /// can be proven from the function's where-clause assumptions.
+    pub(crate) fn verify_universal_outlives(
```
nikomatsakis, line 11: Hmmn, note to self, right now we are doing this on every `prove_goal`, but I think in this setup we could actually move it to run at the end of the fn, since outlives will always propagate up? I guess maybe not if we added some kind of outlives-edge killing.

crates/formality-rust/src/grammar/expr/codegen.rs
```diff
@@ -0,0 +1,229 @@
+#![expect(unused_variables)]
```
nikomatsakis, line 1: This whole file is still a WIP

PR #258, opened 2026-03-24
enable creation of lang item crate

BennoLossin:
For the formality work on field projections, we need to define some custom traits that are always present. My idea was to just create a crate that's implicitly added to each `Program`. In order to do those changes later, I have made the `crates` field private and changed the constructors.

See my other branch on how I added the new lang items crate: https://github.com/BennoLossin/a-mir-formality/tree/field-projections

I spoke with @dingxiangfei2009 and he could also make use of this for the reborrowing project goal.

### DIFF
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -21,6 +21,7 @@ pub mod borrow_check;
 
 mod adts;
 mod coherence;
+mod core_crate;
 mod fns;
 mod impls;
 mod traits;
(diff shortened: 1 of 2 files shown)


crates/formality-rust/src/check/mod.rs
```diff
@@ -16,13 +16,13 @@ pub mod borrow_check;
 /// Check all crates in the program. The crates must be in dependency order
 /// such that any prefix of the crates is a complete program.
 pub fn check_all_crates(program: &Program) -> Fallible<ProofTree> {
-    let Program { crates } = program;
+    let crates = program.crates();
```
nikomatsakis: I prefer to destructure so I know when fields are added. You can add `x: _` if you want to hide something.

crates/formality-rust/src/grammar/program.rs
```diff
@@ -3,13 +3,26 @@ use crate::grammar::{Fallible, TraitId};
 use formality_core::term;
 
 #[term($crates)]
+#[customize(constructors)]
 pub struct Program {
     /// List of all crates.
     /// The last crate in the list is the current crate.
-    pub crates: Vec<Crate>,
+    crates: Vec<Crate>,
```
nikomatsakis: Why not pub?

PR #259, opened 2026-03-24
 feat: add CoreFold::Output type and CoreIntoFold for tuple substitution

### LINKED ISSUE #247: Cleanup: change CoreFold output type to support folding reference tuples
Follow-up cleanup from PR #246:

`CoreFold` currently requires the output type to be `Self`, which prevents `apply(&(&Wcs, &Wcs))` from producing `(Wcs, Wcs)`. Changing this would let `prove_after.rs` collapse two separate `apply` calls into a single tuple destructure:

```rust
// current (two calls)
(let assumptions = c1.substitution().apply(assumptions))
(let goal = c1.substitution().apply(goal))

// desired (one call)
(let (assumptions, goal) = c1.substitution().apply(&(assumptions, goal)))
```

See: https://github.com/rust-lang/a-mir-formality/pull/246#discussion_r2889280008

System625:
Closes #247.

Adds type Output to CoreFold so that container types like Vec<T>, Option<T>, and tuples propagate their inner type's output generically. Implements       
CoreFold for &T with Output = T::Output, allowing apply to accept owned values directly. This collapses two apply calls in prove_after.rs into one:

```
// before
(let assumptions = c1.substitution().apply(assumptions))
(let goal        = c1.substitution().apply(goal))

// after
(let (assumptions, goal) = c1.substitution().apply((assumptions, goal)))
```

AI Disclosure: Used Claude as a learning aid for understanding the CoreFold trait system and iterating on the approach. All code reviewed and tested locally.

Confidence: 8/10 — tests pass and the approach matches what was requested, but still learning the trait system nuances.

Questions: Is the Output = Self bound on the Fold trait alias in declare_language! the right place to constrain it, or should that be looser?

### DIFF
--- a/crates/formality-rust/src/prove/prove/prove/prove_after.rs
+++ b/crates/formality-rust/src/prove/prove/prove/prove_after.rs
@@ -15,7 +15,7 @@ judgment_fn! {
         debug(constraints, goal, assumptions)
 
         (
-            (let (assumptions, goal) = c1.substitution().apply((&assumptions, &goal)))
+            (let (assumptions, goal) = c1.substitution().apply((assumptions, goal)))
             (prove(decls, c1.env(), assumptions, goal) => c2)
             --- ("prove_after")
             (prove_after(decls, c1, assumptions, goal) => c1.seq(c2))


crates/formality-core/src/fold.rs
```diff
@@ -39,109 +43,151 @@ pub trait CoreFold<L: Language>: Sized + CoreVisit<L> {
     }
 }
 
-impl<L: Language, T: CoreFold<L>> CoreFold<L> for Vec<T> {
+impl<L: Language, T: CoreFold<L, Output = T>> CoreFold<L> for Vec<T> {
+    type Output = Self;
```
nikomatsakis: This is in the right direction, but not quite what I had in mind. The idea would be that the `Output` should be...

```rust
impl<L: Language, T: CoreFold<L>> CoreFold<L> for Vec<T> {
    type Output = Vec<T::Output>;
```

...this, combined with an impl of `CoreFold` for `&T` that yields `T::Output`, should allow us to take references in more places. 

I'm not convinced we need this other trait you added.

crates/formality-rust/src/prove/prove/prove/prove_after.rs
```diff
@@ -15,9 +15,8 @@ judgment_fn! {
         debug(constraints, goal, assumptions)
 
         (
-            (let assumptions = c1.substitution().apply(assumptions))
-            (let goal = c1.substitution().apply(goal))
-            (prove(decls, c1.env(), assumptions, goal) => c2)
+            (let (assumptions, goal) = c1.substitution().apply((assumptions, goal)))
+            (prove(decls, c1.env(), &assumptions, &goal) => c2)
```
nikomatsakis: This is a bad sign =) the goal is to have *fewer* ampersands, not more.

nikomatsakis: I'm not sure what the specific examples were from that issue, but I'd like to see them getting cleaner here :)

crates/formality-rust/src/prove/prove/prove/prove_after.rs
```diff
@@ -15,8 +15,7 @@ judgment_fn! {
         debug(constraints, goal, assumptions)
 
         (
-            (let assumptions = c1.substitution().apply(assumptions))
-            (let goal = c1.substitution().apply(goal))
+            (let (assumptions, goal) = c1.substitution().apply((&assumptions, &goal)))
```
lqd: We'd want to avoid these ampersands too, IIUC.

PR #263, opened 2026-03-27
fix: generalize over collection type

nikomatsakis:
The older parsing code allowed `$[v0]` etc to parse as any collection. The newer changes I made hard-coded `Vec`. In opening this PR, I realize it needs a test. Will add later.

### DIFF
--- a/tests/parser-torture-tests/ambiguity_helpers.rs
+++ b/tests/parser-torture-tests/ambiguity_helpers.rs
@@ -40,10 +40,7 @@ fn count_complete_parses<T: CoreParse<crate::ptt::FormalityLang> + std::fmt::Deb
         .into_iter()
         .filter(|p| parse::skip_whitespace(p.text()).is_empty())
         .collect();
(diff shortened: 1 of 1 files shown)


tests/parser-torture-tests/ambiguity_helpers.rs
```diff
@@ -22,14 +24,30 @@ pub enum Amb {
...
     let parses = T::parse(&scope, input).unwrap();
-    parses
+    let complete: Vec<_> = parses
         .into_iter()
         .filter(|p| parse::skip_whitespace(p.text()).is_empty())
-        .count()
+        .collect();
+    let values: Vec<T> = complete.into_iter().map(|p| p.finish().0).collect();
+    assert_eq!(
+        values.len(),
```
lqd, line 45: 😭

nikomatsakis: Why the tears? Note that `values` is used for add'l things.

PR #276, opened 2026-03-28
Convert to judgment groundwork

nikomatsakis:
This PR lays the groundwork to convert the "top-level Rust functions" into judgment functions. I want us to use judgment functions as much as we can because they're easier to build tooling around and they look cool. It'll also make for a bunch of nice "good first issue" bugs.

I'm going to open a follow-up PR that actually *does* the conversion of a particular function, this is just getting things into place where that conversion is possible.

### DIFF
--- a/crates/formality-rust/src/check/adts.rs
+++ b/crates/formality-rust/src/check/adts.rs
@@ -33,7 +33,12 @@ pub(super) fn check_adt(program: &Program, adt: &Adt) -> Fallible<ProofTree> {
 
     proof_tree
         .children
-        .push(super::where_clauses::prove_where_clauses_well_formed(program, &env, &where_clauses, &where_clauses)?);
+        .push(super::where_clauses::prove_where_clauses_well_formed(
+            program,
+            &env,
+            &where_clauses,
+            &where_clauses,
+        )?);
 
     for Variant { name: _, fields } in &variants {
         for Field { name: _, ty } in fields {
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -66,7 +66,9 @@ fn check_crate(program: &Program, c: &Crate) -> Fallible<ProofTree> {
     check_for_duplicate_items(program)?;
 
     for item in items {
-        proof_tree.children.push(check_crate_item(program, item, &id)?);
+        proof_tree
+            .children
+            .push(check_crate_item(program, item, &id)?);
     }
 
     proof_tree.children.push(check_coherence(program, c)?);
@@ -134,7 +136,13 @@ fn prove_goal(
     goal: impl ToWcs + Debug,
 ) -> Fallible<ProofTree> {
     let goal: Wcs = goal.to_wcs();
-    prove_judgment(program, env, assumptions, goal.to_wcs(), crate::prove::prove::prove)
+    prove_judgment(
+        program,
+        env,
+        assumptions,
+        goal.to_wcs(),
+        crate::prove::prove::prove,
+    )
 }
 
 fn prove_judgment<G>(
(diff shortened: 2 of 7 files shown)


issues-draft.json
tiif, line 1: What's the reason of having the json file?

issues-draft.md
tiif, line 1: It seems a bit weird to have the issues draft in the repo 🤔  

maybe we can just open an issue or send it in zulip or dump it in hackmd?

crates/formality-core/src/judgment/proven_set.rs
```diff
@@ -681,6 +681,20 @@ pub enum RuleFailureCause {
...
+        } else if let Some(failed) = e.downcast_ref::<FailedJudgment>() {
+            RuleFailureCause::FailedJudgment(Box::new(failed.clone()))
+        } else {
+            RuleFailureCause::Inapplicable {
+                reason: format!("{e:?}"),
+            }
+        }
+    }
+}
+
```
tiif, line 697: I am not sure the effect of adding this, maybe it will be shown in the next PR?

crates/formality-rust/src/check/mod.rs
```diff
@@ -15,12 +15,12 @@ pub mod borrow_check;
 
 /// Check all crates in the program. The crates must be in dependency order
 /// such that any prefix of the crates is a complete program.
-pub fn check_all_crates(program: &Program) -> Fallible<ProofTree> {
-    let Program { crates } = program;
+pub fn check_all_crates(program: &Crates) -> Fallible<ProofTree> {
+    let Crates { crates } = program;
```
lqd: Do we want to remove the mentions of "program"?

crates/formality-rust/src/check/mod.rs
```diff
@@ -35,8 +35,7 @@ pub fn check_all_crates(program: &Crates) -> Fallible<ProofTree> {
 fn check_current_crate(program: &Crates) -> Fallible<ProofTree> {
     let decls = program.to_prove_decls();
     Check {
-        program,
-        decls: &decls,
+        program: &decls,
     }
     .check()
```
lqd: ok so now we also want to remove mentions of decls?

issues-draft.json
lqd, line 1: stray file?

issues-draft.md
lqd, line 1: here as well?

crates/formality-rust/src/check/adts.rs
```diff
@@ -2,50 +2,49 @@ use std::collections::HashSet;
...
+    let mut env = Env::default();
+
+    let AdtBoundData {
+        where_clauses,
+        variants,
+    } = env.instantiate_universally(binder);
+
+    proof_tree
+        .children
+        .push(super::where_clauses::prove_where_clauses_well_formed(program, &env, &where_clauses, &where_clauses)?);
```
lqd: note to self: why is this taking where clauses twice

crates/formality-core/src/judgment/proven_set.rs
```diff
@@ -859,6 +859,22 @@ pub trait EachProof {
     ) -> Result<(), RuleFailureCause>;
 }
 
+impl EachProof for Fallible<ProofTree> {
```
lqd: this seems to be removed in the next commit 🤔

issues-draft.json
nikomatsakis: oh, I didn't mean to add that file, yeah.

crates/formality-rust/src/check/adts.rs
nikomatsakis: Answer: one is the where-clauses for the assumptions, the other is the where-clauses that need to be proven as WF

crates/formality-core/src/judgment/proven_set.rs
nikomatsakis: ah yeah, I changed strategy on that, because I forgot that I made `(foo() => ())` go through a different path now

crates/formality-rust/src/check/mod.rs
nikomatsakis: yeah, missed a few renames...

Issue #267, opened 2026-03-27
Add type-check test coverage for `Stmt::If`


nikomatsakis:
Parent tracking issue: #265

The [`Stmt::If`](https://github.com/rust-lang/a-mir-formality/blob/1f47788eccc7b9eeb7159a1d0c7453324d6e74a2/crates/formality-rust/src/grammar/expr/mod.rs#L56) variant has borrow-check coverage (e.g., `min_problem_case_3`) but no type-checking coverage in `mir_typeck.rs`.

Add a test that exercises an `if` statement in the surface language. For example, a function that branches on a condition and returns different values from each branch.

Look at the existing borrow-check tests using `if` for the syntax pattern. See the parent tracking issue for instructions on how to add tests.

PR #277, opened 2026-03-28
test: add test for Stmt::If for type-check coverage

xonx4l:
This PR adds a test in `mir_typeck.rs` as discussed in https://github.com/rust-lang/a-mir-formality/issues/267

A new test `if_else()` with a small function is added .

This test covers `Stmt::If` type-check coverage in mir_typeck.rs .

Test via :
cargo test if_else
cargo test

AI disclosure:
No AI was used  .

### DIFF
--- a/src/test/mir_typeck.rs
+++ b/src/test/mir_typeck.rs
@@ -146,6 +146,28 @@ fn if_else() {
     )
 }
 
+// if / else (Stmt::If) with different return types.
+#[test]
+fn if_else_different_return_types() {
+    crate::assert_err!(
+        [
+            crate Foo {
+                fn foo (b: bool) -> u32 {
+                    if b {
+                        return 1 _ u32;
+                    } else {
+                        return 0 _ bool;
+                    }
+                }
+            }
+        ]
+        expect_test::expect![[r#"
+            crates/formality-rust/src/prove/prove/prove/prove_normalize.rs:19:1: no applicable rules for prove_normalize { p: bool, assumptions: {}, env: Env { variables: [], bias: Soundness, pending: [], allow_pending_outlives: true } }
+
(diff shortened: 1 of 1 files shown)


src/test/mir_typeck.rs
```diff
@@ -114,6 +114,24 @@ fn test_cyclic_goto() {
...
+                    if b {
+                        return 1 _ u32;
+                    } else {
+                        return 2 _ u32;
+                    }
+                }
+            }
+        ]
+    )
+}
```
nikomatsakis, line 147: This looks good -- but let's add a test where the two arms have *different* types too!

lqd: I said that as well earlier :3

xonx4l: Added the test with different return types . Thanks for the review @nikomatsakis and @lqd .

nikomatsakis: Oops, missed that lqd

src/test/mir_typeck.rs
```diff
@@ -128,6 +128,46 @@ fn test_ret_true() {
...
+                    if b {
+                        return 1 _ u32;
+                    } else {
+                        return 0 _ bool;
+                    }
+                }
+            }
+        ]
+        expect_test::expect![[r#"
+            crates/formality-rust/src/prove/prove/prove/prove_normalize.rs:19:1: no applicable rules for prove_normalize { p: bool, assumptions: {}, env: Env { variables: [], bias: Soundness, pending: [], allow_pending_outlives: true } }
```
nikomatsakis, line 165: interesting error report. Could be better, but not the fault of this PR.

nikomatsakis (approved):

PR #282, opened 2026-03-28
test: borrow-check test coverage for ExprData::Struct

### LINKED ISSUE #272: Add borrow-check test coverage for `ExprData::Struct`
Parent tracking issue: #265

The [`ExprData::Struct`](https://github.com/rust-lang/a-mir-formality/blob/1f47788eccc7b9eeb7159a1d0c7453324d6e74a2/crates/formality-rust/src/grammar/expr/mod.rs#L165) variant has type-checking coverage (e.g., `test_struct`) but no borrow-check coverage in `borrowck.rs`.

Add borrow-check tests exercising struct construction, e.g.:
- Constructing a struct that borrows from a local
- Accessing a struct field while another field is borrowed

See the parent tracking issue for instructions on how to add tests.

pranavpurankar:
Closes #272

### What does this PR do?
This PR introduces borrow-check coverage for `ExprData::Struct` to `borrowck.rs`. 
Added tests to check construction of a struct & field projections. I included one positive path and two negative paths to ensure the `place_disjoint_from_place` logic correctly catches invalid overlapping accesses.

### Test plan
* Added `struct_disjoint_field_borrows` (`assert_ok!`)
* Added `struct_conflicting_field_borrows` (`assert_err!`)
* Added `struct_construction_with_borrowed_local` (`assert_err!`)
* Added `struct_with_mutable_reference_locks_local` (`assert_err!`)

### Ran
* `UPDATE_EXPECT=1 cargo test` to generate the snapshots for negative tests. `cargo test` passes locally.
*  `cargo test --all` for CI checks
*  `cargo fmt` - final formatting
---
### AI disclosure &  PR context
**AI tools used:** Yes, for research & PR formatting. I used strictly as a "rubber duck" to help map the formal logic of place projections (how `p.x` relates to `p`) into the new `a-mir-formality` syntax during the `WIP.md` planning phase. Code was written and tested locally.

**Confidence level:** I feel good about it. The negative tests correctly trigger the expected `place_disjoint_from_place`, `!outlived_by_loan`,`write-indirect`,`loan_cannot_outlive` rule failures.

**Questions for mentors:** The disjointness checks passed. For future NLL test cases involving deeper nested structs, does the `Expr` grammar require any special handling for deref projections (e.g., (*p).x) or does it treat them identically to standard field access in the borrow checker?

### DIFF
--- a/src/test/borrowck.rs
+++ b/src/test/borrowck.rs
@@ -1197,3 +1197,116 @@ fn call_mut_under_shared_borrow() {
 
     )
 }
+
+/// mutably borrowing two distinct fields of a struct -> assert_ok!
+#[test]
+fn struct_disjoint_field_borrows() {
+    crate::assert_ok!(
+        [
+            crate Foo {
+                struct Point { x: u32, y: u32 }
+                fn foo() -> u32 {
+                    exists<'r0, 'r1, 'r2, 'r3> {
+                        let p: Point = Point { x: 0 _ u32, y: 0 _ u32 };
+                        let b1: &mut 'r0 u32 = &mut 'r1 p.x;
+                        let b2: &mut 'r2 u32 = &mut 'r3 p.y;
+                        *b1 = 1 _ u32;
+                        *b2 = 2 _ u32;
+                        return 0 _ u32;
+                    }
+                }
+            }
+        ]
+    )
+}
+
+/// accessing a field while it is already mutably borrowed -> borrow error
+#[test]
+fn struct_conflicting_field_borrows() {
+    crate::assert_err!(
+        [
+            crate Foo {
+                struct Point { x: u32, y: u32 }
+                fn foo() -> u32 {
+                    exists<'r0, 'r1> {
+                        let p: Point = Point { x: 0 _ u32, y: 0 _ u32 };
+                        let b1: &mut 'r0 u32 = &mut 'r1 p.x;
+                        p.x = 1 _ u32;
+                        return *b1;
+                    }
+                }
+            }
+        ]
+        expect_test::expect![[r#"
+            the rule "borrow of disjoint places" at (nll.rs) failed because
+              condition evaluated to false: `place_disjoint_from_place(&loan.place, &access.place)`
+                &loan.place = p : Point . x : u32
+                &access.place = p : Point . x : u32
+
+            the rule "loan_cannot_outlive" at (nll.rs) failed because
+              condition evaluated to false: `!outlived_by_loan.contains(&lifetime.upcast())`
+                outlived_by_loan = {?lt_1, ?lt_2}
+                &lifetime.upcast() = ?lt_1
+
+            the rule "write-indirect" at (nll.rs) failed because
+              pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p`
+
+            the rule "write-indirect" at (nll.rs) failed because
+              pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p : Point . x`"#]]
+    )
+}
+
+// constructing a struct reading a local variable that is mutably borrowed -> borrow error
+#[test]
+fn struct_construction_with_borrowed_local() {
+    crate::assert_err!(
+    [
+        crate Foo {
+            struct Wrapper {
+                value: u32,
+            }
+            fn foo() -> u32 {
+                exists<'r0, 'r1> {
+                    let v1: u32 = 22 _ u32;
+                    let v2: &mut 'r0 u32 = &mut 'r1 v1;
+                    let w: Wrapper = Wrapper { value: v1 };
+                    return *v2;
+                }
+            }
+        }
+    ]
+    expect_test::expect![[r#"
+        the rule "borrow of disjoint places" at (nll.rs) failed because
+          condition evaluated to false: `place_disjoint_from_place(&loan.place, &access.place)`
+            &loan.place = v1 : u32
+            &access.place = v1 : u32
+
+        the rule "loan_cannot_outlive" at (nll.rs) failed because
+          condition evaluated to false: `!outlived_by_loan.contains(&lifetime.upcast())`
+            outlived_by_loan = {?lt_1, ?lt_2}
+            &lifetime.upcast() = ?lt_1"#]]
+    )
+}
+
+/// placing a mutable reference inside a struct -> locks the underlying local variable
+#[test]
+fn struct_with_mutable_reference_locks_local() {
+    crate::assert_err!(
+        [
+            crate Foo {
+                struct Wrapper<'a> {
+                    value: &mut 'a u32,
+                }
+                fn foo() -> u32 {
+                    exists<'r0> {
+                        let v1: u32 = 0 _ u32;
+                        let w: Wrapper<'r0> = Wrapper::<'r0> { value: &mut 'r0 v1 };
+                        v1 = 1 _ u32;
+                        return *(w.value);
+                    }
+                }
+            }
+        ]
+        expect_test::expect!["crates/formality-rust/src/prove/prove/prove/prove_wf.rs:14:1: no applicable rules for prove_wf { goal: ?lt_0, assumptions: {}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }"]
+    )
+}
--- a/src/test/borrowck.rs
+++ b/src/test/borrowck.rs
@@ -1307,6 +1307,7 @@ fn struct_with_mutable_reference_locks_local() {
                 }
             }
         ]
+        // FIXME(#304) -- This does not look like the error message we expect.
         expect_test::expect!["crates/formality-rust/src/prove/prove/prove/prove_wf.rs:14:1: no applicable rules for prove_wf { goal: ?lt_0, assumptions: {}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }"]
     )
 }


WIP.md
```diff
@@ -0,0 +1,22 @@
+# WIP: Borrow-check coverage for `ExprData::Struct` (#272)
+
+## Goals
```
nikomatsakis: can you delete this file in the final commit

src/test/borrowck.rs
```diff
@@ -998,3 +998,93 @@ fn write_to_borrowed_before_zero_iteration_loop() {
...
+              pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p`
+
+            the rule "write-indirect" at (nll.rs) failed because
+              pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p : Point . x`"#]]
+    )
+}
+
+/// Test that constructing a struct reading a local variable that is mutably borrowed fails.
+#[test]
+fn struct_construction_with_borrowed_local() {
```
nikomatsakis, line 1261: let's add a test like

```rust
crate Foo {
    struct Wrapper<'a> {
        value: &mut 'a u32
    }

    fn foo() {
        let v1 = 0 _ u32;
        let w = Wrapper { value: &mut v1 };
        v1; // Error
        w.value;
     }
}
```

nikomatsakis (changes requested):

WIP.md
pranavpurankar: Yes, it has been removed.

src/test/borrowck.rs
pranavpurankar: I've added test for this : shows error when underlying local is accessed while the struct hold a mut ref.

src/test/borrowck.rs
```diff
@@ -999,6 +999,23 @@ fn write_to_borrowed_before_zero_iteration_loop() {
...
+                struct Point { x: u32, y: u32 }
+                fn foo() -> u32 {
+                    exists<'r0, 'r1, 'r2, 'r3> {
+                        let p: Point = Point { x: 0 _ u32, y: 0 _ u32 };
+                        let b1: &mut 'r0 u32 = &mut 'r1 p.x;
+                        let b2: &mut 'r2 u32 = &mut 'r3 p.y;
+                        *b1 = 1 _ u32;
+                        *b2 = 2 _ u32;
+                        return 0 _ u32;}}}])
+}
```
nikomatsakis:
```suggestion
/// mutably borrowing two distinct fields of a struct -> assert_ok!
#[test]
fn struct_disjoint_field_borrows() {
    crate::assert_ok!(
        [
            crate Foo {
                struct Point { x: u32, y: u32 }
                fn foo() -> u32 {
                    exists<'r0, 'r1, 'r2, 'r3> {
                        let p: Point = Point { x: 0 _ u32, y: 0 _ u32 };
                        let b1: &mut 'r0 u32 = &mut 'r1 p.x;
                        let b2: &mut 'r2 u32 = &mut 'r3 p.y;
                        *b1 = 1 _ u32;
                        *b2 = 2 _ u32;
                        return 0 _ u32;
                    }
                }
            }
        ]
    )
}
```

src/test/borrowck.rs
```diff
@@ -1023,6 +1040,38 @@ fn call_pass_ref() {
...
+                        outlived_by_loan = {?lt_1, ?lt_2}
+                        &lifetime.upcast() = ?lt_1
+
+                    the rule "write-indirect" at (nll.rs) failed because
+                      pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p`
+
+                    the rule "write-indirect" at (nll.rs) failed because
+                      pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p : Point . x`"#]]
+                )
+}
```
nikomatsakis:
```suggestion
fn struct_conflicting_field_borrows() {
    crate::assert_err!(
    [
        crate Foo {
            struct Point { x: u32, y: u32 }
            fn foo() -> u32 {
                exists<'r0, 'r1> {
                    let p: Point = Point { x: 0 _ u32, y: 0 _ u32 };
                    let b1: &mut 'r0 u32 = &mut 'r1 p.x;
                    p.x = 1 _ u32;
                    return *b1;
                }
            }
        }
    ]
    expect_test::expect![[r#"
        the rule "borrow of disjoint places" at (nll.rs) failed because
          condition evaluated to false: `place_disjoint_from_place(&loan.place, &access.place)`
            &loan.place = p : Point . x : u32
            &access.place = p : Point . x : u32

        the rule "loan_cannot_outlive" at (nll.rs) failed because
          condition evaluated to false: `!outlived_by_loan.contains(&lifetime.upcast())`
            outlived_by_loan = {?lt_1, ?lt_2}
            &lifetime.upcast() = ?lt_1

        the rule "write-indirect" at (nll.rs) failed because
          pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p`

        the rule "write-indirect" at (nll.rs) failed because
          pattern `TypedPlaceExpressionData::Deref(place_loaned_ref)` did not match value `p : Point . x`"#]]
    )
}
```

src/test/borrowck.rs
```diff
@@ -1085,3 +1166,27 @@ fn call_mut_under_shared_borrow() {
...
+    crate::assert_err!(
+        [
+            crate Foo {
+                struct Wrapper<'a> {
+                    value: &mut 'a u32,
+                }
+                fn foo() -> u32 {
+                    exists<'r0> {
+                        let v1: u32 = 0 _ u32;
+                        let w: Wrapper<'r0> = Wrapper { value: &mut 'r0 v1 };
```
nikomatsakis:
```suggestion
                        let w: Wrapper<'r0> = Wrapper::<'r0> { value: &mut 'r0 v1 };
```

I think you might need this.

src/test/borrowck.rs
```diff
@@ -1085,3 +1166,27 @@ fn call_mut_under_shared_borrow() {
...
+                        v1 = 1 _ u32;
+                        let ptr: &mut 'r0 u32 = w.value;
+                        return *ptr;
```
nikomatsakis:
```suggestion
                        return *(w.value);
```

nikomatsakis: let's try that

src/test/borrowck.rs
```diff
@@ -1085,3 +1166,27 @@ fn call_mut_under_shared_borrow() {
...
+                    }
+                }
+            }
+        ]
+        expect_test::expect!["crates/formality-rust/src/prove/prove/prove/prove_wf.rs:14:1: no applicable rules for prove_wf { goal: ?lt_0, assumptions: {}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }"]
```
nikomatsakis, line 1311: Hmm, wait, this doesn't look like the correct error. I think it's a bit of an unrelated bug, might be worth landing as-is, or I can talk you through how to fix it.

Let's land the changes above first.

nikomatsakis (changes requested):

src/test/borrowck.rs
nikomatsakis: So I *think* what's happening here is that the well-formedness check is failing on `let w: Wrapper<'r0>` and it's probably because `exists` needs to add assumptions into the env that each of the new lifetime inference variables is well-formed:

https://github.com/rust-lang/a-mir-formality/blob/e70e68f3a33a9650942c8d12c7e31abe522c3075/crates/formality-rust/src/check/borrow_check/nll.rs#L278-L286

Basically the `(borrow_check_block(env, assumptions, state, block, places_live_on_exit) => state)` would need to be changed to union the `assumptions` with `subst.iter().map(|v| v.well_formed())` or something like that.

nikomatsakis:
@pranavpurankar (needs rebase too it looks like)

src/test/borrowck.rs
pranavpurankar: Thank you!

src/test/borrowck.rs
pranavpurankar: @nikomatsakis Thank you for the direction currently I understand that `env.instantiate_existentially` creates the new lifetime variables, but I need some time to study the borrow check block and nll.rs carefully before coming to a conclusion.

I will raise an issue after I’ve had a chance to communicate on Zulip. I’m little cautious now bcz recent changes broke the CI. First I'll study this based on your provided direction before moving forward.

nikomatsakis: Let's discuss, I'd love to fix this before merging

nikomatsakis: Actually I take that back -- we can open a follow-up issue and land this first

pranavpurankar: @nikomatsakis Sounds good. I understood the initial left brain urge to fix this (_Let's discuss, I'd love to fix this before merging_) but it may create a **ripple effect**. I tried to fix it locally based on your suggestion but I need to study this more. I have a direction now as you suggested and Ill try to fix it later.

I agree that this PR has served its purpose. We'll open up a follow-up issue for the `nll.rs` well-formed bug. This PR is ready to merge, I think.

nikomatsakis: Opened https://github.com/rust-lang/a-mir-formality/pull/282#discussion_r3019048262

nikomatsakis: https://github.com/rust-lang/a-mir-formality/issues/304

nikomatsakis:
```suggestion
        // FIXME(#304) -- This does not look like the error message we expect.
        expect_test::expect!["crates/formality-rust/src/prove/prove/prove/prove_wf.rs:14:1: no applicable rules for prove_wf { goal: ?lt_0, assumptions: {}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }"]
```

src/test/borrowck.rs
```diff
@@ -1197,3 +1197,117 @@ fn call_mut_under_shared_borrow() {
...
+            crate Foo {
+                struct Wrapper<'a> {
+                    value: &mut 'a u32,
+                }
+                fn foo() -> u32 {
+                    exists<'r0> {
+                        let v1: u32 = 0 _ u32;
+                        let w: Wrapper<'r0> = Wrapper::<'r0> { value: &mut 'r0 v1 };
+                        v1 = 1 _ u32;
+                        return *(w.value);
```
pranavpurankar, line 1305: parser currently cannot distinguish between *(w.value) and (*w).value 
PR #331 fixes it

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/282/7d82d1b09827b39cd605bceb2b68e12bd9ebc01b..55ebe8d236f227d8dd6407e8748e8ddf31a37df7)*

PR #283, opened 2026-03-28
Fix left-recursive parsing, add reject

nikomatsakis:
## What does this PR do?

Two distinct fixes, both motivated by ambiguities found when upgrading data-model

### Parse all "seeds" in left-recursive grammars

Before this PR, when parsing a left-recursive grammar like Perm = leaf | Perm Perm, the fixed-point loop only keeps the longest parse as the seed for the next iteration. So parsing leaf leaf leaf would find `Apply(Leaf, Apply(Leaf, Leaf))` (right-associative) but never `Apply(Apply(Leaf, Leaf), Leaf)` (left-associative), because the intermediate `Apply(Leaf, Leaf)` was never tried as a seed. The fix is to use all accumulated values as seeds, not just the longest one.

Each round now snapshots the full set into the stack entry (type-erased as `Box<dyn Any>` because we are using thread-local storage that must be monomorphic). As a drive-by improvement, also removes the unsafe that was previously needed for raw-pointer-based seed passing.

### Reject attribute

The `#[reject]` attribute is a separate feature to enable the parser to resolve inter-type ambiguities. In the Dada grammar, both types and permissions can be applied. This means that e.g. `given given i32` can be parsed as `(given given) i32` (i.e., an `Apply` perm of `(given given)` applied to a type of `i32`) or as `given (given i32)` (i.e., an `ApplyPerm` type with `given` applied to another `ApplyPerm` type with `given`). These ought to both be equivalent in the Dada type rules, but we do need to pick one when parsing.

The `#[reject]` lets you do point-wise adjustments, complementing the existing precedence mechanism, basically saying: reject parses where this is true. In this case, we can say that on `ApplyPerm` we reject a `Perm` that is itself an `Apply`, thus pushing us to prefer `given (given i32)`.

(An alternative would be to reject a `Ty` that is an `ApplyPerm`, which would prefer `(given given) i32`.)

## Disclosure and PR context

**AI tools used:** written in collaboration with `pi`, the problem doc is in the first commit (and removed at the end)

**Confidence level:** Feel good about it

**Review depth:** I've read over the code some, but I'd like to do another review, that's why I'm opening it as a Draft PR

**Testing:** Tested, there a number of new tests

**Questions for mentors:** None. But I'm realizing, I ought to include some docs!

### DIFF
--- a/crates/formality-macros/src/parse.rs
+++ b/crates/formality-macros/src/parse.rs
@@ -788,7 +788,7 @@ fn wrap_with_reject(
     let reject_clauses: Vec<TokenStream> = rejects
         .iter()
         .map(|reject| reject_clause(variant, reject))
-        .collect::< syn::Result<Vec<_>>>()?;
+        .collect::<syn::Result<Vec<_>>>()?;
 
     // OR all reject clauses together
     let combined = if reject_clauses.len() == 1 {


tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    Id(PermId),
+
+    #[grammar($v0 $v1)]
+    Apply(Arc<Perm>, Arc<Perm>),
+}
+
+formality_core::id!(PermId);
+
+/// Count the number of complete parses (no remaining input) for a type.
+fn count_complete_parses<T: CoreParse<crate::ptt::FormalityLang>>(input: &str) -> Vec<T>
```
nikomatsakis: We should rename this to `all_parses`

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    let results = count_complete_parses::<Perm>("leaf leaf");
+    assert_eq!(
+        results.len(),
+        1,
+        "expected exactly 1 parse for `leaf leaf`, got {}: {results:#?}",
+        results.len()
+    );
+
+    let term: Perm = crate::ptt::term("leaf leaf");
+    expect_test::expect![[r#"
```
nikomatsakis, line 64: We should apply `expect_test::expect!` directly to `results`

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    assert_eq!(
+        results.len(),
+        1,
+        "expected 1 parse due to right-associative bias (should be 2 when fixed), \
+         got {}: {results:#?}",
+        results.len()
+    );
+
+    // The single result should be the right-associative one:
+    let term: Perm = crate::ptt::term("leaf leaf leaf");
```
nikomatsakis: As above, silly to have a separate call to `term` here

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    // With the bias, we expect fewer than 5 parses.
+    // Document whatever the current (biased) behavior is.
+    assert!(
+        results.len() < 5,
+        "expected fewer than 5 parses due to bias, got {}: {results:#?}",
+        results.len()
+    );
+
+    // At minimum the right-associative parse should be present:
+    let term: Perm = crate::ptt::term("leaf leaf leaf leaf");
```
nikomatsakis: As above

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+#[test]
+fn ty_simple_perm_application() {
+    let results = count_complete_parses::<Ty>("leaf Data");
+    assert_eq!(
+        results.len(),
+        1,
+        "expected 1 parse for `leaf Data`, got {}: {results:#?}",
+        results.len()
+    );
+}
```
nikomatsakis, line 215: we should use expect-test here

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+            results.len(),
+            results
+        );
+    }
+
+    // Just assert we get at least 1 parse
+    assert!(
+        !results.is_empty(),
+        "expected at least 1 parse for `leaf x Data`"
+    );
```
nikomatsakis: we're going to fix this test in a later commit, right?

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    let results = count_complete_parses::<Ty>("leaf x y Data");
+    eprintln!(
+        "deeper cross-type ambiguity: {} parse(s): {:#?}",
+        results.len(),
+        results
+    );
+    assert!(
+        !results.is_empty(),
+        "expected at least 1 parse for `leaf x y Data`"
+    );
```
nikomatsakis: we should use expect-test

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -0,0 +1,325 @@
...
+    for val in &complete_values {
+        eprintln!("  {val:?}");
+    }
+    eprintln!("Incomplete parses ({}):", incomplete_values.len());
+    for (val, rem) in &incomplete_values {
+        eprintln!("  {val:?} (remaining: {rem:?})");
+    }
+
+    // Build the expected trees for comparison
+    let right_assoc = Perm::Apply(
```
nikomatsakis: we should just use expect-test

crates/formality-macros/src/parse.rs
```diff
@@ -736,3 +759,106 @@ fn wrap_bindings(bindings: &[BindingInfo], tail: TokenStream, current_index: usi
...
+                } else {
+                    checks.push(quote! {
+                        matches!(DowncastFrom::downcast_from(&#name), Some(#pattern))
+                    });
+                }
+            }
+        }
+    }
+
+    // If trailing `..`, remaining fields are implicitly `_` (skip)
```
nikomatsakis: Hmm. I guess we are not going to be testing that the complete set of fields was given, etc. i.e., if people fail to use `..` when they should, it works fine. I'm not sure that this matters, I'm tempted to just say... it's fine. We can just remove `..` and say we only match on the fields that you name.

nikomatsakis: But then we should update the docs and remove support for `..`

tests/parser-torture-tests/reject.rs
```diff
@@ -0,0 +1,278 @@
...
+/// "Non-skipped fields are AND'd"
+///
+/// If there are no non-skipped fields, the conjunction of zero things is
+/// vacuously true. But that would mean #[reject(_, _)] rejects everything,
+/// which seems wrong. I think the intent is: if ALL fields are wildcards,
+/// there's no effective constraint, so __rejected = false.
+///
+/// Actually the AND of zero predicates is `true`, but the reject fires when
+/// the conjunction is `true`. So `#[reject(_, _)]` = always reject.
+/// That's a valid (if strange) usage. Let's test that it rejects everything.
```
nikomatsakis: This is a very weird comment. But indeed this `_, _` mean everything is rejected. You can strip this comment to something much simpler!

crates/formality-core/src/parse/parser/left_recursion.rs
```diff
@@ -245,67 +311,60 @@ where
...
-                    tracing::trace!(
-                        "precedence_valid = {}",
-                        precedence_valid,
-                    );
-                    if !precedence_valid {
+                    // Filter seeds by precedence compatibility with the current variant.
+                    let valid_seeds: Set<SuccessfulParse<'_, T>> = all_seeds
+                        .into_iter()
+                        .filter(|seed| {
+                            let precedence_valid = match current_precedence.associativity {
```
nikomatsakis, line 349: This is some subtle logic. I remember thinking all this through, but I think it'd be helpful to add a comment here giving an example as to why this is correct. Right now it's not very obvious to me!

crates/formality-core/src/parse/parser/left_recursion.rs
```diff
@@ -398,7 +457,9 @@ where
     //
     // Now we return sets of successful results rather than a single best result.
     // The fixed-point iteration accumulates all successful parses across rounds,
-    // using the best (longest) parse as the seed for left-recursion reuse.
+    // using ALL accumulated values as seeds for left-recursion reuse. This fixes
+    // the right-associative bias where only the longest parse was used as a seed,
+    // preventing intermediate-length parses from composing further.
```
nikomatsakis: This comment is "documenting the journey, not the destination". The reader will have no idea what bias is being referred to, as it no longer exists! Remove the last sentence you goof.

tests/parser-torture-tests/ambiguity.rs
```diff
@@ -59,23 +59,21 @@ fn reduce_reduce_ambig() {
...
-                a,
-                b,
-            ),
-            OneId(
-                c,
-            ),
-        )
-    "#]]
-    .assert_debug_eq(&term);
+    // With all-seeds iteration, the parser now finds multiple complete parses.
```
nikomatsakis: This comment is "documenting the journey, not the destination". Remove the word "now".

tests/parser-torture-tests/ambiguity.rs
```diff
@@ -59,23 +59,21 @@ fn reduce_reduce_ambig() {
...
+        .into_iter()
+        .filter(|p| parse::skip_whitespace(p.text()).is_empty())
+        .collect();
+
+    // At least 2 complete parses due to TwoRr's left-recursion finding multiple trees.
+    assert!(
+        complete.len() >= 2,
+        "expected multiple ambiguous parses for `a b c`, got {}: {complete:#?}",
+        complete.len()
+    );
```
nikomatsakis: include `expect_test` and change the assert to assert exactly the right number, not just >= 2.

tests/parser-torture-tests/left_recursion_seed.rs
```diff
@@ -62,27 +62,30 @@ fn left_recursion_seed_shadow() {
...
+    );
+
+    let (value, _) = complete.into_iter().next().unwrap().finish();
+    let expected = E::ThenEnd(std::sync::Arc::new(E::Id(
+        Id::new("x"),
+    )));
+    assert_eq!(
+        value, expected,
+        "expected ThenEnd(Id(x))"
+    );
```
nikomatsakis: use expect_test here

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -93,28 +93,14 @@ fn two_leaves() {
...
+    // Both left- and right-associative parses should be present.
     assert_eq!(
         results.len(),
-        1,
-        "expected 1 parse due to right-associative bias (should be 2 when fixed), \
+        2,
+        "expected 2 parses (both associativities) for `leaf leaf leaf`, \
          got {}: {results:#?}",
         results.len()
     );
```
nikomatsakis: add an exepct-test call on `results` here

tests/parser-torture-tests/left_recursion_seed_bias.rs
```diff
@@ -127,29 +113,14 @@ fn three_leaves_both_associativities() {
...
-        results.len() < 5,
-        "expected fewer than 5 parses due to bias, got {}: {results:#?}",
+    // Catalan number C_3 = 5 distinct binary trees with 4 leaves.
+    assert_eq!(
+        results.len(),
+        5,
+        "expected 5 parses (Catalan C_3) for `leaf leaf leaf leaf`, \
+         got {}: {results:#?}",
         results.len()
     );
```
nikomatsakis: add an expect-test call on the results vec here

tests/parser-torture-tests/reject.rs
```diff
@@ -65,28 +67,45 @@ where
 /// Without #[reject], it produces 2 (the cross-type ambiguity from wip.md).
 #[test]
 fn reject_cross_type_ambiguity() {
-    let results = count_complete_parses::<Ty>("leaf x Data");
+    let results = complete_parses::<Ty>("leaf x Data");
```
nikomatsakis, line 72: ...oh, like we renamed this already...

tests/parser-torture-tests/reject.rs
```diff
@@ -208,33 +267,16 @@ pub enum Ty6 {
...
-/// If there are no non-skipped fields, the conjunction of zero things is
-/// vacuously true. But that would mean #[reject(_, _)] rejects everything,
-/// which seems wrong. I think the intent is: if ALL fields are wildcards,
-/// there's no effective constraint, so __rejected = false.
-///
-/// Actually the AND of zero predicates is `true`, but the reject fires when
-/// the conjunction is `true`. So `#[reject(_, _)]` = always reject.
-/// That's a valid (if strange) usage. Let's test that it rejects everything.
+/// `#[reject(_, _)]` — all fields are wildcards, so the AND of zero predicates
+/// is vacuously true, meaning every parse through this variant is rejected.
```
nikomatsakis: oh, you fixed this goofy comment already

lqd:
I read through it to unblock you (but didn’t have the time to review per se).

PR #284, opened 2026-03-28
Check that `let` uses a well-formed type

mattheww:
a-mir-formality currently accepts the `let` below, although `S2<S1>` isn't well formed.

```
crate Crate1 {
    trait Trait1 {}
    struct S1 {}
    struct S2<T> where T: Trait1 {
        t: T,
    }
    fn f() -> () {
        let s2: S2<S1>;
    }
}
```

I think this is the natural fix, basically just adding `ty.well_formed()` to the "Let" case of `borrow_check_statement()`.

## Disclosure and PR context

**AI tools used:**

No

**Confidence level:**

I think it's right, want confirmation

**Testing:**

Adds one `assert_ok` and one `assert_err` testcase

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -174,6 +174,7 @@ judgment_fn! {
         debug(state, statement, places_live_on_exit, assumptions, env)
 
         (
+            (prove_ty_is_wf(env, assumptions, state, ty) => state)
             // The variable being declared is not yet live before this `let`,
             // so remove it from places_live_on_exit before checking the initializer (if any).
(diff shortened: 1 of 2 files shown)


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -174,6 +174,7 @@ judgment_fn! {
         debug(state, statement, places_live_on_exit, assumptions, env)
 
         (
+            (prove_ty_is_wf(env, assumptions, state, ty) => state)
```
nikomatsakis, line 177: Yes, I think this is right -- I think we have some WF checks elsewhere, but for the fn body, this seems like the right place to me.

nikomatsakis (approved):

PR #285, opened 2026-03-29
Adjust grammar for `trusted` functions to prevent early consumtion of the `trusted` keyword

liolin:
## What does this PR do?

The new Expr syntax has changed the format for trusted functions. However, this new format is not always parsed correctly.
If there is no where clause (and therefore no `where` keyword), the `trusted` keyword is consumed by the parser when checking whether the next string is the `where` keyword. By surrounding the `trusted` keyword with curly braces, the parser no longer consumes `trusted` when figuring out, if there are where clauses.

## Disclosure and PR context

**AI tools used:** No

**Confidence level:** The code works. However, I'm not sure if this is the correct solution to the problem, as it seems like a bug in the parser implementation.

**Review depth:** I reviewed closely and understand it well (only a few lines...)

**Testing:** I ran full test suite (cargo test --all)

**Questions for mentors:** Would the proposed syntax also be OK for you? Would you like the parser to at least throw an error?

### DIFF
--- a/crates/formality-rust/src/grammar/fns.rs
+++ b/crates/formality-rust/src/grammar/fns.rs
@@ -33,7 +33,7 @@ pub enum MaybeFnBody {
 
 #[term]
 pub enum FnBody {
-    #[grammar(trusted;)]
+    #[grammar({trusted})]
     TrustedFnBody,
 
     #[cast]
--- a/crates/formality-rust/src/test.rs
+++ b/crates/formality-rust/src/test.rs
@@ -154,7 +154,7 @@ fn test_parse_trusted_fn() {
     let r: Program = term(
         "[
             crate core {
-              fn run() -> () trusted;
+              fn run() -> () {trusted}
             }
         ]",
     );
--- a/src/test/functions.rs
+++ b/src/test/functions.rs
@@ -36,7 +36,7 @@ fn lifetime() {
                 fn one_lt_arg<'a, T>(v0: &'a T) -> ()
                 where
                     T: 'a, // FIXME(#202): Implied bounds should not have to be explicit
-                trusted;
+                {trusted}
             }
         ]
     )


crates/formality-rust/src/grammar/fns.rs
```diff
@@ -33,7 +33,7 @@ pub enum MaybeFnBody {
 
 #[term]
 pub enum FnBody {
-    #[grammar(trusted;)]
+    #[grammar({trusted})]
```
nikomatsakis, line 36: I guess this works now because blocks always expect lists of statements -- this used to be ambiguous. I'm not sure I understand the *bug* exactly, but I kind of like this syntax better anyway.

liolin: > I'm not sure I understand the bug exactly

I have a minimal example, which should hopefully make things clearer:

```rs
#[test]
fn test_parser() {
    #[term($:where $foo $bar)]
    struct T {
        foo: Option<crate::grammar::ValueId>,
        bar: Bar,
    }

    #[term]
    enum Bar {
        #[grammar(trusted)]
        Trusted,
    }

    let r: T = crate::rust::term("trusted");
    expect_test::expect![[r#"
      T {
          foo: None,
          bar: Trusted,
      }
    "#]]
    .assert_debug_eq(&r);
}
```

When I run `RUST_LOG=trace RUST_BACKTRACE=0 cargo test -- test_parser`, I get the following output:

```
running 1 test
 TRACE enter<formality_rust::test::test_parser::{{closure}}::T>(scope=Scope { bindings: [] }, text="trusted")
 TRACE min_precedence_level = 0
 nonterminal name="T", scope=Scope { bindings: [] }, text="trusted"
   variant name="T::T", variant_precedence=Precedence { level: 18446744073709551614, associativity: Both }
     expect_keyword self=ActiveVariant { precedence: Precedence { level: 18446744073709551614, associativity: Both }, scope: Scope { bindings: [] }, start_text: "trusted", current_text: "trusted", is_committed: true }, expected="where"
       identifier_like_string
           0ms TRACE return=Ok("trusted")
         0ms TRACE return=Err({ParseError { text: "trusted", message: "expected `where`", backtrace: [ParseFrame { name: "T", remaining_len: 7 }] }})
       0ms TRACE enter<formality_rust::test::test_parser::{{closure}}::Bar>(scope=Scope { bindings: [] }, text="")
       0ms TRACE min_precedence_level = 0
     nonterminal name="Bar", scope=Scope { bindings: [] }, text=""
       variant name="Bar::Trusted", variant_precedence=Precedence { level: 18446744073709551614, associativity: Both }
         expect_keyword self=ActiveVariant { precedence: Precedence { level: 18446744073709551614, associativity: Both }, scope: Scope { bindings: [] }, start_text: "", current_text: "", is_committed: true }, expected="trusted"
           identifier_like_string
               0ms TRACE return=Err({ParseError { text: "", message: "unexpected end of input", backtrace: [ParseFrame { name: "T", remaining_len: 7 }, ParseFrame { name: "Bar", remaining_len: 0 }] }})
             0ms TRACE return=Err({ParseError { text: "", message: "expected `trusted`", backtrace: [ParseFrame { name: "T", remaining_len: 7 }, ParseFrame { name: "Bar", remaining_len: 0 }] }})
       0ms TRACE parsing failed: no variants were able to consume a single token
     0ms TRACE error: ParseError { text: "", message: "Bar expected", backtrace: [ParseFrame { name: "T", remaining_len: 7 }, ParseFrame { name: "Bar", remaining_len: 0 }] }
 TRACE parsing failed with 1 errors `{ParseError { text: "", message: "Bar expected", backtrace: [ParseFrame { name: "T", remaining_len: 7 }, ParseFrame { name: "Bar", remaining_len: 0 }] }}`
test test::test_parser ... FAILED

failures:

---- test::test_parser stdout ----

thread 'test::test_parser' (514415) panicked at crates/formality-rust/src/test.rs:211:16:
called `Result::unwrap()` on an `Err` value: × Bar expected
   ╭────
 1 │ trusted
   · ▲
   · ╰── while parsing T
   ╰────
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test::test_parser

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 30 filtered out; finished in 0.00s

error: test failed, to rerun pass `--lib`
```

It seems, that the `expect_keyword` and `identifier_like_string` consume the input string, meaning that other parsing options no longer have access to the original string. The line `nonterminal name="Bar", scope=Scope { bindings: [] }, text=""` indicates, that the remaining string is the empty string, when it should be "trusted".

This behavior is not necessarily wrong, but it is not what I expected.

PR #287, opened 2026-03-30
convert top-level functions to judgment_fn

nikomatsakis:
## What does this PR do?

Convert the top-level functions (`check_all_crates`, `check_crate`, `check_crate_item`) to judgment functions. Avoids the need to manually construct proof trees.

This will serve as an exemplar for converting similar functions elsewhere.

## Disclosure and PR context

**AI tools used:** not this time.

**Confidence level:** feel good about it

**Review depth:** I reviewed closely and understand it well

**Testing:** `UPDATE_EXPECT=1 cargo test --all --workspace`

**Questions for mentors:** none

### DIFF
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -1,6 +1,6 @@
 #![allow(dead_code)]
 
-use std::{collections::VecDeque, fmt::Debug};
+use std::fmt::Debug;
 
 use crate::prove::prove::{is_definitely_not_proveable, Constraints, Env, Program};
 use crate::rust::Visit;
@@ -9,7 +9,7 @@ use crate::{
     prove::ToWcs,
 };
 use anyhow::{anyhow, bail};
-use formality_core::{judgment::ProofTree, ProvenSet, Set};
+use formality_core::{judgment::ProofTree, judgment_fn, ProvenSet, Set};
 
 use adts::check_adt;
 use coherence::check_coherence;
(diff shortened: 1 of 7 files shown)


crates/formality-rust/src/check/mod.rs
```diff
@@ -19,64 +19,52 @@ use traits::check_trait;
...
+    /// Check that the current crate `c` is valid,
+    /// given the complete program (which includes dependencies).
+    fn check_crate(
+        program: Program,
+        c: Crate,
+    ) => () {
+        debug(c, program)
+
+        (
+            (check_for_duplicate_items(program) => ())
```
nikomatsakis, line 57: This should probably be moved to `check_all_crates`

nikomatsakis: Meh, keep it here I guess.

crates/formality-rust/src/check/mod.rs
```diff
@@ -19,64 +19,52 @@ use traits::check_trait;
...
+
+        (
+            // Check that all crates up to and including crate #i are valid.
+            // Crate #i will be considered the "current crate".
+            (for_all(i in 0..crates.len())
+                (let program = crates.prefix(i).to_prove_decls())
+                (check_crate(program, &crates.crates[i]) => ()))
+            ------------------------------------------------------------ ("check all prefixes")
+            (check_all_crates(crates) => ())
+        )
```
lqd, line 43: Sorry, this is pre-existing but could you remind me why we set things up this way, checking prefixes instead of splitting per-crate and per-program concerns independently? We may have documented this elsewhere already, I'm not sure.

I'm just now wondering how essential this is, and how it will behave on more real-world looking code, under our efforts to expand our test inputs.

nikomatsakis: Hmm. There's no particular reason. It kind of models what cargo does, but I think we could *just* check the leaf crate conceivably, or have both modes.

nikomatsakis: I'm going to land this PR but I think we should talk this out, might be a good simplification, though I still think i'd like most tests to test all crates

lqd: I surely am misunderstanding the setup, and being confused about what we are checking multiple times when looking at a program representing a prefix of this list of crates.

PR #301, opened 2026-03-31
Convert check_trait to judgment function

JojoFlex1:
AI tools used: Claude (Anthropic)
Confidence level: This is an early draft, I want feedback
Review depth: I read the code and understand the general direction and what output is expected, but have questions
Testing: cargo test 109/120 passing. 4 tests fail because where_clauses checking and env handling inside judgment_fn needs guidance
Questions for mentors: How should I handle prove_where_clauses_well_formed inside judgment_fn since it requires &Env which needs instantiate_universally which requires &mut env?

What's working:
- check_trait is now a judgment_fn
- check_trait_items_have_unique_names stays as a regular helper function as noted in the tracking issue

What's not working yet:
- 4 tests failing because I couldn't figure out how to handle prove_where_clauses_well_formed inside judgment_fn , it requires &Env but instantiate_universally requires &mut env which judgment_fn closures don't allow

I have read the docs and what output is expected . Happy to fix whatever needs changing.

### DIFF
--- a/crates/formality-rust/src/check/traits.rs
+++ b/crates/formality-rust/src/check/traits.rs
@@ -1,52 +1,25 @@
-use crate::grammar::{
-    AssociatedTy, AssociatedTyBoundData, Fn, Trait, TraitBoundData, TraitItem, WhereClause,
-};
+use crate::grammar::{AssociatedTy, Trait, TraitBoundData, TraitItem};
 use crate::grammar::{CrateId, Fallible};
-use crate::prove::prove::{Env, Program};
+use formality_core::{judgment_fn, judgment::ProofTree, Set};
 use anyhow::bail;
-use fn_error_context::context;
-use formality_core::{judgment::ProofTree, Set};
+use crate::prove::prove::{Env, Program};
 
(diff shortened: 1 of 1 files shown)


nikomatsakis (changes requested):
@JojoFlex1 --

> Confidence level: This is an early draft, I want feedback

OK! It looks like this is a good start but you've removed most of the checks, so that won't do.

> Questions for mentors: How should I handle `prove_where_clauses_well_formed` inside `judgment_fn` since it requires `&Env` which needs `instantiate_universally` which requires `&mut env`?

Right -- this method is going to need to be replaced. Really what we *ought* to do is to refactor all those methods to take `&self` -- they should clone internally and return the clone.

Maybe a good starting point would be to make a fresh PR just doing that refactoring.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/301/5e810150f799c26a15c45b56f86b55ff327131c7..6a455803e7a59bcdbea3385a7a76b9d31f6eb279)*

PR #305, opened 2026-04-04
add test to ensure wf of references

BennoLossin:
## What does this PR do?

Fixes a problem in the well-formedness rule of references. This code is successfully checked by a-mir-formality:
```rust
crate core {
    trait Trait {}
    struct A<X> where X: Trait {}
    fn invalid<'a, X>(x: &'a A<X>) -> ()
    where
        X: 'a,
    {}
}
```
This is obviously wrong, since `X` must implement `Trait` for `A<X>` to be wf.

## Disclosure and PR context

**AI tools used:** None

**Confidence level:** High in that this is an issue, a bit unsure on my solution, I don't know if I composed the constraints correctly.

**Testing:** `cargo test --all` and added a new test to catch invalid references (which failed before my modifications)

### DIFF
--- a/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
+++ b/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
@@ -6,7 +6,7 @@ use formality_core::{judgment_fn, Downcast, ProvenSet};
 
 use crate::prove::prove::{
     decls::Program,
-    prove::{combinators::for_all, prove_after::prove_after, prove_wc::prove_wc},
+    prove::{combinators::for_all, prove_after::prove_after},
 };
 
 use super::{constraints::Constraints, env::Env};
@@ -34,7 +34,8 @@ judgment_fn! {
         (
             // `&'a T` is well-formed if `T: 'a`
             (let (lt, ty) = parameters.downcast_err::<(Lt, Ty)>()?)
-            (prove_wc(decls, env, assumptions, Relation::outlives(ty, lt)) => c)
+            (prove_wf(decls, env, assumptions, ty) => c)
+            (prove_after(decls, c, assumptions, Relation::outlives(ty, lt)) => c)
             --- ("references")
             (prove_wf(decls, env, assumptions, RigidTy { name: RigidName::Ref(_), parameters }) => c)
         )
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -8,6 +8,7 @@ mod consts;
 mod decl_safety;
 mod functions;
 mod mir_typeck;
+mod references;
 mod well_formed_trait_ref;
 
 #[test]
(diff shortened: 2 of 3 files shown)


tests/invalid_ref.rs
tiif, line 1: This test is not really about normalization, could you rename the test and add a comment about why this test exist? Also it will be nice to have a pass test too. :)

crates/formality-rust/src/prove/prove/prove/prove_wf.rs
```diff
@@ -34,9 +34,11 @@ judgment_fn! {
         (
             // `&'a T` is well-formed if `T: 'a`
             (let (lt, ty) = parameters.downcast_err::<(Lt, Ty)>()?)
-            (prove_wc(decls, env, assumptions, Relation::outlives(ty, lt)) => c)
+            (prove_wf(decls, env, assumptions, ty) => ty_c)
+            (let (assumptions, rel) = ty_c.substitution().apply((assumptions, Relation::outlives(ty, lt))))
+            (prove_wc(decls, ty_c.env(), assumptions, rel) => lt_c)
```
tiif: You might be able to use ``prove_after`` to do the "apply assumption then prove goal" dance.

nikomatsakis: Yes, as @tiif suggested..

```suggestion
            (prove_after(decls, ty_c, assumptions, Relation::outlives(ty, lt)) => lt_c)
```

...should work. My preference is to do `=> c` and then shadow it again here (vs `ty_c` and `lt_c`). The reason is that it is a bug if you use the wrong constraint, basically, it has to flow linearly.

nikomatsakis (changes requested):

tests/invalid_ref.rs
BennoLossin: Oh the name of the test was a copy paste x)

Created https://github.com/rust-lang/a-mir-formality/issues/312 to add a comment with the context.

BennoLossin: ~~I also tried to move the test, but tests in `src/test` aren't actually run under `cargo test` or `cargo test --all`?~~

~~I'll just leave it where it is and let you handle the test restructuring.~~

Edit: :facepalm: just found the `mod.rs`

crates/formality-rust/src/prove/prove/prove/prove_wf.rs
BennoLossin: I looked at `prove_after` in order to understand how the constraints are combined and then I didn't use it, because I thought that `prove_wc` would do something different than `prove_after`, since the latter takes a `Wcs` and the former handles a `Wc`. Turns out you do have conversions set up, so this should work exactly as intended. (I'm pretty new to a-mir-formality :)

PR #306, opened 2026-04-04
field projection lang items

BennoLossin:
This is nowhere near ready, just using this to make it easier for folks to give feedback.

(no AI was used for this)

### DIFF
--- a/crates/formality-rust/src/check/core_crate.rs
+++ b/crates/formality-rust/src/check/core_crate.rs
@@ -2,6 +2,8 @@ use formality_core::term;
 
 use crate::grammar::{Crate, CrateId, CrateItem};
 
+mod field_projections;
+
 #[term($*items)]
 struct LangItems {
     items: Vec<CrateItem>,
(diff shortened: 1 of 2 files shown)


crates/formality-rust/src/check/core_crate.rs
```diff
@@ -1,8 +1,329 @@
-use crate::grammar::Crate;
+use crate::grammar::{Crate, CrateId};
+
+/// Parses `a-mir-formality` pseudo-Rust syntax into a `Vec` of crate items.
+macro_rules! lang_items {
+    // # Macro States
```
nikomatsakis: I've nothing against this macro, but I wonder...why not just parse to a `Vec<CrateItem>`?

nikomatsakis: Or make a:

```rust
#[term($*items)]
pub struct CoreCrate {
    items: Vec<CrateItem>
}
```

and parse that?

BennoLossin: :facepalm: the reason for me wanting this macro is that the lang items get syntax highlighting. I never considered that `Vec<T>` also implemented `Term` when `T` did... (since that led to bad errors in my personal parsing projects :)

It does work with your suggestion.

BennoLossin:
AFAIK we don't have bors yet, so this probably won't work

@bors r=@nikomatsakis

Issue #289, opened 2026-03-30
Convert `check_trait` to a judgment function


nikomatsakis:
Parent tracking issue: #288

Convert [`check_trait`](https://github.com/rust-lang/a-mir-formality/blob/1f47788eccc7b9eeb7159a1d0c7453324d6e74a2/crates/formality-rust/src/check/traits.rs#L12) from a regular Rust function to a judgment function.

Currently it: opens the trait binder, proves where-clauses are well-formed, and checks each trait item.

Also convert the related helpers:
- `check_trait_item` — dispatches on `TraitItem::Fn` vs `TraitItem::AssociatedTy`
- `check_fn_in_trait` — delegates to `check_fn`
- `check_associated_ty` — proves where-clauses WF

See the parent tracking issue for context and examples.

nikomatsakis:
I think was completed

PR #307, opened 2026-04-06
Convert check_trait to judgment function

JojoFlex1:
Hey @nikomatsakis , i redid the failing tests and changes the instantiate to self , all test re passing and i also created a new pr for this . What i did is Change Env::instantiate_universally from &mut self to &self, returning (Env, T) — a new environment with the fresh universal variables in scope, plus the instantiated value. This makes it callable from within judgment_fn closures which only allow &Env.
All call sites updated to destructure the tuple. All tests passing. I hope this was the right way . Waiting for feedback.
#289

### DIFF
--- a/crates/formality-rust/src/check/traits.rs
+++ b/crates/formality-rust/src/check/traits.rs
@@ -18,7 +18,7 @@ judgment_fn! {
             (let (env, bound_data) = env.instantiate_universally(&binder.explicit_binder))
             (let TraitBoundData { where_clauses, trait_items } = bound_data)
             (check_trait_items_have_unique_names(&trait_items) => ())
-            (super::where_clauses::prove_where_clauses_well_formed(&program, &new_env, &where_clauses, &where_clauses) => ())
+            (super::where_clauses::prove_where_clauses_well_formed(program, env, where_clauses, where_clauses) => ())
             ------------------------------------------------------------ ("check trait")
             (check_trait(program, env, t, crate_id) => ())
         )


crates/formality-rust/src/check/traits.rs
```diff
@@ -1,52 +1,28 @@
...
+        program: Program,
+        env: Env,
+        t: Trait,
+        crate_id: CrateId,
+    ) => () {
+        debug(program, t, crate_id)
+
+        (
+            (let Trait { safety: _, id: _, binder } = &t)
+            (let (new_env, bound_data) = env.instantiate_universally(&binder.explicit_binder))
```
nikomatsakis:
```suggestion
            (let (env, bound_data) = env.instantiate_universally(&binder.explicit_binder))
```

I'm weird this way but I prefer to shadow in cases where it would be a bug to use the old name.

crates/formality-rust/src/check/traits.rs
```diff
@@ -1,52 +1,28 @@
...
+            (let TraitBoundData { where_clauses, trait_items } = bound_data)
+            (check_trait_items_have_unique_names(&trait_items) => ())
+            (super::where_clauses::prove_where_clauses_well_formed(&program, &new_env, &where_clauses, &where_clauses) => ())
```
nikomatsakis:
```suggestion
            (super::where_clauses::prove_where_clauses_well_formed(program, env, where_clauses, where_clauses) => ())
```

I think you will find these are not needed (and I changed to use the shadowed name)

crates/formality-rust/src/prove/prove/prove/env.rs
```diff
@@ -274,17 +274,18 @@ impl Env {
 
     /// Given a bound value `<X..> v`, returns `v` with `X...` replaced with fresh universal variables.
     /// Modifies `self` to bring those variables into scope.
-    pub fn instantiate_universally<T>(&mut self, b: &Binder<T>) -> T
+    pub fn instantiate_universally<T>(&self, b: &Binder<T>) -> (Env, T)
```
nikomatsakis, line 277: Nice!

nikomatsakis (changes requested):
This looks great @JojoFlex1 ! Left two nits. I'll just apply my patches.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/307/517ef31dad2cc8f9544833940c27302a56d85710..2f1e99ff3bb0a0fcf34e98385b9d651adcc70e50)*

PR #310, opened 2026-04-06
Reject moves out of references

### LINKED ISSUE #297: Reject moves out of references
You cannot move a value out from behind a reference (`&T` or `&mut T`). Currently formality allows this.

In rustc this is error E0507: "cannot move out of `*r` which is behind a shared/mutable reference." This is a property of the *place* being behind a reference, not about active loans — it applies even if there are no other borrows.

Once `Copy` is implemented, `Copy` types *can* be read through shared references (it's a copy, not a move). So the full rule is: "if the place is behind a reference, the type must be `Copy`; otherwise it's an error."

**Ignored tests** (`#[ignore]` in `src/test/borrowck.rs`):

- `move_out_of_shared_ref` — `let y: u32 = *r` where `r: &u32`
- `move_out_of_mut_ref` — `let y: u32 = *r` where `r: &mut u32`

System625:
## What does this PR do?

- Add `prove_place_is_movable` judgment function that walks a `TypedPlaceExpr` and succeeds only when the place is not behind a reference. It handles `Local` and `Field` cases; `Deref` has no matching rules yet (no owned deref types like `Box`), so all derefs through `&`/`&mut` naturally fail.
- Wire `prove_place_is_movable` into the "place" rule in `borrow_check_expr` at the existing FIXME
- Un-ignore `move_out_of_shared_ref` and `move_out_of_mut_ref` tests and update their expected errors

 Closes #297

## Disclosure and PR context

**AI tools used:** Claude Code helped with codebase exploration, implementation approach, and drafting the judgment function

**Confidence level:** High — the approach was discussed with @nikomatsakis on Zulip and follows his guidance to "think positively" with a `prove_place_is_movable` judgment rather than a negative `place_is_behind_ref` check

**Review depth:** I read and understand all changes

**Testing:** `move_out_of_shared_ref` and `move_out_of_mut_ref` both pass. `cargo check` passes. Full `cargo test --all` hit OOM on Windows locally — relying on CI for full suite.

**Questions for mentors:** None

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -432,15 +432,7 @@ judgment_fn! {
             (access_permitted(env, assumptions, state, Access::new(AccessKind::Read, place), places_live_on_exit) => state)
             // FIXME(#296): also need to track that the place has been moved from
             (prove_place_is_movable(env, assumptions, state, place) => state)
-            ------------------------------------------------------------ ("place-move")
-            (borrow_check_expr(env, assumptions, state, ExprData::Place(place), places_live_on_exit) => (&place.ty, state))
-        )
-
-        (
-            (borrow_check_place_expr(env, assumptions, state, place) => (place, state))
-            (access_permitted(env, assumptions, state, Access::new(AccessKind::Read, place), places_live_on_exit) => state)
-            (prove_ty_is_copy(env, assumptions, state, &place.ty) => state)
-            ------------------------------------------------------------ ("place-copy")
+            ------------------------------------------------------------ ("place")
             (borrow_check_expr(env, assumptions, state, ExprData::Place(place), places_live_on_exit) => (&place.ty, state))
         )
 
@@ -750,8 +742,10 @@ judgment_fn! {
 }
 
 judgment_fn! {
-    /// Prove that a type is `Copy`. For now, scalar types and references are
-    /// considered `Copy`. This will eventually be replaced by a proper trait check.
+    /// Prove that a type is `Copy`. For now, scalar types and shared references
+    /// are considered `Copy`.
+    ///
+    /// FIXME: replace with a proper trait check once lang-items are available.
     fn prove_ty_is_copy(
         env: TypeckEnv,
         assumptions: Wcs,
@@ -790,21 +784,36 @@ judgment_fn! {
         debug(env, assumptions, state, place)
 
         (
-            (if let TypedPlaceExpressionData::Local(_) = place.data())
             ------------------------------------------------------------ ("local")
-            (prove_place_is_movable(_env, _assumptions, state, place) => state)
+            (prove_place_is_movable(
+                _env,
+                _assumptions,
+                state,
+                TypedPlaceExpressionData::Local(_),
+            ) => state)
         )
 
         (
-            (if let TypedPlaceExpressionData::Field(prefix, _) = place.data())
             (prove_place_is_movable(env, assumptions, state, prefix) => state)
             ------------------------------------------------------------ ("field")
+            (prove_place_is_movable(
+                env,
+                assumptions,
+                state,
+                TypedPlaceExpressionData::Field(prefix, _),
+            ) => state)
+        )
+
+        (
+            (prove_ty_is_copy(env, assumptions, state, &place.ty) => state)
+            ------------------------------------------------------------ ("copy")
             (prove_place_is_movable(env, assumptions, state, place) => state)
         )
 
-        // For a deref, we would need to prove the prefix type is an "owned deref"
-        // (e.g., Box). Since no owned deref types exist yet, this rule has no
-        // matching cases and derefs through references naturally fail.
+        // For a deref of a non-Copy type, we would need to prove the prefix type
+        // is an "owned deref" (e.g., Box). Since no owned deref types exist yet,
+        // this rule has no matching cases and derefs through references naturally
+        // fail.
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -740,6 +749,65 @@ judgment_fn! {
     }
 }
 
+judgment_fn! {
+    /// Prove that a type is `Copy`. For now, scalar types and references are
+    /// considered `Copy`. This will eventually be replaced by a proper trait check.
```
nikomatsakis: We should open a FIXME for this, it is blocked somewhat on @BennoLossin's work towards lang-items.

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -740,6 +749,65 @@ judgment_fn! {
...
+            ------------------------------------------------------------ ("local")
+            (prove_place_is_movable(_env, _assumptions, state, place) => state)
+        )
+
+        (
+            (if let TypedPlaceExpressionData::Field(prefix, _) = place.data())
+            (prove_place_is_movable(env, assumptions, state, prefix) => state)
+            ------------------------------------------------------------ ("field")
+            (prove_place_is_movable(env, assumptions, state, place) => state)
+        )
```
nikomatsakis, line 811: I have a style / diagnostics nit here. If this judgment is invoked with a `Place::Deref`, then this rule will show up as a 'possible match', which isn't (imo) desired. The fix is to either add an `!` after the rules that say that this rule "almost applied":

```rust
        (
            (if let TypedPlaceExpressionData::Field(prefix, _) = place.data())! // <-- note the `!`
            (prove_place_is_movable(env, assumptions, state, prefix) => state)
            ------------------------------------------------------------ ("field")
            (prove_place_is_movable(env, assumptions, state, place) => state)
        )
```

The `!` is a way to say "if this `if let` doesn't match, don't consider this rule as a possible match". It's useful when you are "refining" the match expressions in some sense. But my preferred fix is just to move that stuff into the patterns, like so:

```suggestion
        (
            (prove_place_is_movable(env, assumptions, state, prefix) => state)
            ------------------------------------------------------------ ("field")
            (prove_place_is_movable(
                env,
                assumptions,
                state,
                TypedPlaceExpressionData::Field(prefix, _),
             ) => state)
        )
```

src/test/borrowck.rs
```diff
@@ -423,7 +422,18 @@ fn move_out_of_shared_ref() {
...
 
-        expect_test::expect![[""]]
+        expect_test::expect![[r#"
+            the rule "scalar" at (nll.rs) failed because
+              pattern `(RigidTy { name: RigidName::ScalarId(_), .. }, state)` did not match value `(Datum, flow_state([scope(none, None, {}, None, [], []), scope(none, None, {}, None, [], []), scope(some(U(2)), None, {}, None, [(x, Datum), (r, &?lt_1 Datum)], [x : Datum, r : &?lt_1 Datum])], point_flow_state({pending_outlives(?lt_2, ?lt_1)}, {loan(?lt_2, x : Datum, shared)}), {}, {}))`
+
+            the rule "shared-ref" at (nll.rs) failed because
+              pattern `(RigidTy { name: RigidName::Ref(RefKind::Shared), .. }, state)` did not match value `(Datum, flow_state([scope(none, None, {}, None, [], []), scope(none, None, {}, None, [], []), scope(some(U(2)), None, {}, None, [(x, Datum), (r, &?lt_1 Datum)], [x : Datum, r : &?lt_1 Datum])], point_flow_state({pending_outlives(?lt_2, ?lt_1)}, {loan(?lt_2, x : Datum, shared)}), {}, {}))`
+
+            the rule "field" at (nll.rs) failed because
```
nikomatsakis: (this is the output I am referring to -- it's like "duh the rule for fields didn't apply to a deref")

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -740,6 +749,65 @@ judgment_fn! {
...
+    /// A place is movable if it is not behind a reference. When a `Deref` is
+    /// encountered, the deref'd type must be an "owned deref" (e.g., `Box`).
+    /// Since we don't model owned deref types yet, all derefs through
+    /// references (`&T` / `&mut T`) will cause this judgment to fail.
+    fn prove_place_is_movable(
+        env: TypeckEnv,
+        assumptions: Wcs,
+        state: FlowState,
+        place: TypedPlaceExpr,
+    ) => FlowState {
```
nikomatsakis, line 783: I think we should just add a third rule to this like

```
(
    (prove_ty_is_copy(env, state, &place.ty) => state)
    --- ("copy")
    (prove_place_is_movable(env, assumptions, state, place) => state)
)
```

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -430,8 +430,17 @@ judgment_fn! {
...
+            (prove_place_is_movable(env, assumptions, state, place) => state)
+            ------------------------------------------------------------ ("place-move")
+            (borrow_check_expr(env, assumptions, state, ExprData::Place(place), places_live_on_exit) => (&place.ty, state))
+        )
+
+        (
+            (borrow_check_place_expr(env, assumptions, state, place) => (place, state))
+            (access_permitted(env, assumptions, state, Access::new(AccessKind::Read, place), places_live_on_exit) => state)
+            (prove_ty_is_copy(env, assumptions, state, &place.ty) => state)
+            ------------------------------------------------------------ ("place-copy")
```
nikomatsakis: ...and then we can delete this rule

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -740,6 +741,82 @@ judgment_fn! {
...
+        (
+            (prove_ty_is_copy(env, assumptions, state, &place.ty) => state)
+            ------------------------------------------------------------ ("copy")
+            (prove_place_is_movable(env, assumptions, state, place) => state)
+        )
+
+        // For a deref of a non-Copy type, we would need to prove the prefix type
+        // is an "owned deref" (e.g., Box). Since no owned deref types exist yet,
+        // this rule has no matching cases and derefs through references naturally
+        // fail.
```
nikomatsakis, line 816: Side note to think about -- I think it'd be useful to add rules that are intentionally omitted. It'd be nice to have a `(fail "...")` rule that indicates why this is not accepted. Let's do this as a follow-up, @System625, I'll open an issue.

nikomatsakis: Done, #315

PR #311, opened 2026-04-07
Formality to rust

liolin:
## What does this PR do?

Generate valid Rust code from formality code (not complete yet) with the `RustBuilder` struct. The `test_util` module provides functions, macros and structs to check formality code against rustc.

## Disclosure and PR context

**AI tools used:** No

**Confidence level:** Feel good about it, but feedback is always welcome.

**Review depth:** I reviewed closely and understand it well (because it is written by me)

**Testing:** I ran full test suite (cargo test --all)

**Questions for mentors:** None


crates/formality-rust/src/grammar/rust_builder/fns.rs
```diff
@@ -0,0 +1,51 @@
...
+        let data = function.binder.peek();
+        let input_args = data
+            .input_args
+            .iter()
+            .map(|a| self.pretty_print_type(&a.ty))
+            .collect::<Result<Vec<_>, _>>()?
+            .join(", ");
+        let output_arg = self.pretty_print_type(&data.output_ty)?;
+        // TODO: Where clauses
+        // self.print_where(&data.where_clauses)?;
```
nikomatsakis: should we open an issue for this?

crates/formality-rust/src/grammar/rust_builder/fns.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        let body = match &data.body {
+            MaybeFnBody::NoFnBody => ";".into(),
+            MaybeFnBody::FnBody(fn_body) => format!(" {}", self.build_fn_body(fn_body)),
+        };
+
+        Ok(format!("fn {id}({input_args}) -> {output_arg}{body}"))
```
nikomatsakis: I expect generics -- I'm thinking that it might be nice to build up a data structure representing "surface Rust" and then write impl `Display` for that data structure....or I guess we kinda already have that...? It feels like there's a level in between

crates/formality-rust/src/grammar/rust_builder/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    pub fn variable_name(
+        &self,
+        core_variable: &CoreVariable<crate::FormalityLang>,
+    ) -> Fallible<String> {
+        self.ctx.variable_name(core_variable)
+    }
+
+    pub fn build_crates(&mut self, crates: &Crates) -> Fallible<Vec<String>> {
```
nikomatsakis: I think it'd be better for the output of this to be a map or something, i.e., the source for each crate, so we can generate a cargo workspace

crates/formality-rust/src/grammar/rust_builder/traits_and_impls.rs
```diff
@@ -0,0 +1,327 @@
...
+        );
+    }
+
+    #[test]
+    fn simple_trait_with_associated_ty2() {
+        crate::assert_rust!(
+            [
+                crate Foo {
+                    trait Write {
+                        type Error<T, K>: [Sized, Bar] where T: Read, K: Write;
```
nikomatsakis: Subtle point -- I think this should map to `T: ?Sized` in surface Rust

crates/formality-rust/src/grammar/rust_builder/traits_and_impls.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    #[test]
+    fn where_is_implemented() {
+        fn t(term: crate::grammar::Trait) -> String {
+            RustBuilder::default().build_trait(&term).unwrap()
+        }
+
+        crate::assert_rust2!(
+            [trait Foo where T: Bar {}],
```
nikomatsakis: I'm a bit confused here, why is this not `trait Foo<T>`?

crates/formality-rust/src/grammar/rust_builder/traits_and_impls.rs
```diff
@@ -0,0 +1,327 @@
...
+            t,
+            "trait Foo<T> where T: Bar { }"
```
nikomatsakis: i.e., this output looks different

crates/formality-rust/src/grammar/rust_builder/traits_and_impls.rs
```diff
@@ -0,0 +1,327 @@
...
+                }
+            ],
+            impl !Write for Bar { }
+        );
+    }
+
+    #[test]
+    fn where_is_implemented() {
+        fn t(term: crate::grammar::Trait) -> String {
+            RustBuilder::default().build_trait(&term).unwrap()
```
nikomatsakis: I don't quite get what we are testing here

nikomatsakis (changes requested):
first pass review! didn't quite finish. Curious to hear your thoughts! this is exciting!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/311/6a4d031757ad6da6931d6f43d2a41b9efa384c9d..109b6b942c28b92e0028e7fe84e81f1814d061de)*

crates/formality-rust/src/test_util.rs
```diff
@@ -0,0 +1,161 @@
...
+            file_name: file_name.into(),
+            input: input.into(),
+            result: None,
+        };
+
+        let compiler_result =
+            rustc_driver::catch_fatal_errors(|| -> rustc_interface::interface::Result<()> {
+                run_compiler(&args, &mut compiler);
+                Ok(())
+            });
```
lqd: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/311/6a4d031757ad6da6931d6f43d2a41b9efa384c9d..109b6b942c28b92e0028e7fe84e81f1814d061de)*

is the plan to still have a driver even though we are generating rust code? are we hooking into internal structures that would prevent us from e.g. only parsing the compiler/cargo stdout/stderr output?

as soon as we're dealing with multiple crates, directly using the driver will be harder, we'll have to use it via wrapper and invoke it via cargo. if we're also not using any driver features, it will also tie us to a rustc revision, and nightly (minirust is also forcing our hand, but we could avoid having to use a nightly to generate the rust code, and could put minirust under a feature requiring nightly, etc)

crates/formality-rust/src/grammar/rust_builder/fns.rs
liolin: This is not required. In another local branch I already implemented it.

crates/formality-rust/src/grammar/rust_builder/mod.rs
liolin: I will adjust accordingly.

crates/formality-rust/src/test_util.rs
liolin: I was also thinking about how to compile generated Rust code. The current approach also has some drawbacks (always printing to stderr, without the possibility to capture the output) which invoking rustc using `std::process::Command` would not have.

I will adjust the invocation accordingly

crates/formality-rust/src/grammar/rust_builder/traits_and_impls.rs
liolin: I will clean up these test code

crates/formality-rust/src/grammar/rust_builder/fns.rs
liolin: Adding data structures that represent surface Rust would add a lot of structs that are similar to the structs in `formality-rust`. An initial idea I had was to implement `Display` for the structs in `formality-rust`. However, this is not possible for all data types because I need a way to convert core variables, which are represented as indexes, to user facing names. And this process requires information that is not encoded in the struct.

I will think about that until next week's sync meeting and hopefully come up with a good solution.

crates/formality-rust/src/test_util.rs
nikomatsakis: I think my preference would be to emit a cargo workspace and run `cargo build.`

lqd: Agreed. It should simplify the code and process, and scale to all the levels of precision we'd need.

liolin:
## What does this PR do?

Generate valid Rust code from formality code (not complete yet) with the `RustBuilder` struct. The `test_util` module provides functions, macros and structs to check formality code against rustc.

When running `cargo test` with the environment variable set `FORMALITY_RUN_RUSTC=1`, the `assert_ok` test cases in `a-mir-formality` are run against rustc.


## Disclosure and PR context

**AI tools used:** Yes, opencode with GPT-5.3-Codex. Used opencode to rewrite the existing code base to the current one. Some adjustments were made by hand.

**Confidence level:** Feel good about it. I like it much

**Review depth:** I reviewed closely and understand it well.

**Testing:** I ran full test suite (cargo test --all)

**Questions for mentors:** Would like to discuss, if the approach with the intermediate representation is the right way for this problem.

crates/formality-rust/src/to_rust/mod.rs
```diff
@@ -0,0 +1,432 @@
...
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct RustBuilder {
+    ctx: NameContext,
+}
+
+impl RustBuilder {
+    pub fn with_binder<T: CoreFold<crate::FormalityLang, Output = T>, R>(
```
nikomatsakis, line 181: Nit: I don't think we need `Output = T` here. I'd rather see `T::Output` below? But it may be needed for some of the impls, if so, ignore me.

crates/formality-rust/src/to_rust/mod.rs
```diff
@@ -0,0 +1,432 @@
...
+        Ok(syntax::Attr::Feature(name.to_owned()))
+    }
+
+    /// Lowers generic parameters and `where` clauses for a binder.
+    ///
+    /// Set `skip_first` when lowering generics for a trait
+    /// declaration.  In trait declarations, the first parameter
+    /// represents `Self` and must not be treated as a regular generic
+    /// parameter.
+    pub fn lower_generics_for_binder(
```
nikomatsakis, line 262: I see that almost every call (maybe every call) to `with_binder` is followed by a call to this. I kind of expected the two to be merged -- e.g., 

```rust
with_named_bindings(
    &t.binder,
    op: impl Fn(&BoundData) -> &[WhereClause],
    skip_first: bool,
) -> ...
```

...but I think it's fine as is, maybe something to consider later.

crates/formality-rust/src/to_rust/test_util.rs
```diff
@@ -0,0 +1,143 @@
...
+        tmp_dir.disable_cleanup(true);
+    }
+    let crates = term(input);
+    let result = check_workspace(&crates, tmp_dir.path()).unwrap();
+    assert!(!result.status.success());
+
+    let stderr = String::from_utf8_lossy(&result.stderr);
+    for code in codes {
+        assert!(stderr.find(*code).is_some());
+    }
```
nikomatsakis, line 99: I'm debating how to manage this. I think my preference would be this:

* extract from rustc (perhaps using json output?) the error code + the error message e.g. `"E123: blah blah blah"`.
* create a `BTreeSet` of those strings.
* have the user of macro pass in a `expect_test::expect![]` as `expect`.
* invoke `expect.assert_eq(&actual_output)`.

This will be more informative than just the error codes and also easy to update.

src/lib.rs
```diff
@@ -48,6 +48,9 @@ pub fn main() -> anyhow::Result<()> {
 macro_rules! assert_ok {
     ($input:tt) => {{
         let _ = $crate::test_program_ok(stringify!($input)).expect("expected program to pass");
+        if $crate::run_rustc() {
+            formality_rust::assert_rustc_success!($input);
```
nikomatsakis, line 52: I'd like at least one set of tests that ALWAYS run rustc before we land this. And also some that do so with errors.

nikomatsakis (changes requested):
Overall, this looks very good. I'd be happy to land if we had *some* tests in CI. 

I also have one suggestion however for how to make the test output more meaningful and easy to update.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/311/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..2e8feecd6bec175712b3b294a2c21e47e92663e4)*

lqd:
I'm not enthused about the env var setup. I'd like to understand the rationale, we can discuss it in today's meeting :)

nikomatsakis:
I assumed that was for testing

nikomatsakis (approved):

nikomatsakis:
Documenting our discussion today: we're going to merge this and then have a follow-up PR to address the test structure, moving away from env variables towards documenting rustc error message and expected result.

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 channel = "nightly-2025-12-01"
-components = ["rustfmt"]
+components = ["rustfmt", "llvm-tools", "rustc-dev", "rust-src"]
```
tiif, line 3: I didn't read through the PR, but I noticed the build get noticeably slow after adding these? May I know why we need all these especially ``llvm-tools``?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/311/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..2e8feecd6bec175712b3b294a2c21e47e92663e4)*

liolin: Sorry, this is an oversight from my initial idea to use private rustc crates. They require `llvm-tools`. I create a PR to fix this.

lqd: Will be fixed by https://github.com/rust-lang/a-mir-formality/pull/337

PR #321, opened 2026-04-09
Flatten Ty, Lt, Const Arc wrappers into enums, moving Arc to use sites in Parameter

System625:
## What does this PR do?

Experiment for #318: Flattens `Ty`/`TyData`, `Lt`/`LtData`, `Const`/`ConstData` Arc-wrapper structs into direct enums, moving `Arc` to use sites in `Parameter` (i.e., `Parameter::Ty(Arc<Ty>)` instead of wrapping Arc inside each type).

This is the first PR in a series, starting with the core types to validate the approach before tackling the rest:

- [x] **PR 1 (this PR):** `Ty`, `Lt`, `Const`
- [ ] **PR 2:** `Wc`/`WcData` (recursive — `ForAll` and `Implies` variants need `Arc`)
- [ ] **PR 3:** `WhereClause`/`WhereClauseData`, `WhereBound`/`WhereBoundData`
- [ ] **PR 4:** `Expr`/`ExprData`, `PlaceExpr`/`PlaceExprData` (heaviest blast radius)

### Test plan
- [x] `cargo build` compiles cleanly (only pre-existing warnings)
- [x] `cargo test --all` — all 269 tests pass
- [x] `cargo fmt --all -- --check` passes

<details>

**AI disclosure.**

* The AI tool authored large parts of the code

**Confidence level.**

* Early draft, want feedback

**Questions for reviewers.**

- Is the temporary `pub type TyData = Ty;` alias the right migration strategy, or should we do a full rename of all `TyData::` references to `Ty::` in this PR?
- For `PredicateTy::ForAll(Binder<Arc<Ty>>)`, is `Arc` the right indirection here, or would `Box` be more appropriate since it's not a shared reference?

</details>

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -1020,7 +1020,7 @@ judgment_fn! {
                 // If `'0: T` then `'0` must hold for entire fn body...
                 Parameter::Ty(_) => false,
 
-                Parameter::Lt(lt) => match &**lt {
+                Parameter::Lt(lt) => match lt.as_ref() {
                     // If `'0: 'static` then `'0` must hold for entire fn body...
                     Lt::Static => false,
 
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -759,13 +759,13 @@ judgment_fn! {
         debug(env, assumptions, state, place)
 
         (
-            (if let TyData::RigidTy(RigidTy { name: RigidName::ScalarId(_), .. }) = place.ty.data())
+            (if let Ty::RigidTy(RigidTy { name: RigidName::ScalarId(_), .. }) = &place.ty)
             ------------------------------------------------------------ ("scalar-copy")
             (access_kind_for_place_use(_env, _assumptions, state, _place) => (AccessKind::Read, state))
         )
 
         (
-            (if let TyData::RigidTy(RigidTy { name: RigidName::Ref(RefKind::Shared), .. }) = place.ty.data())
+            (if let Ty::RigidTy(RigidTy { name: RigidName::Ref(RefKind::Shared), .. }) = &place.ty)
             ------------------------------------------------------------ ("shared-ref-copy")
             (access_kind_for_place_use(_env, _assumptions, state, _place) => (AccessKind::Read, state))
         )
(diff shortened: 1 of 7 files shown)


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -988,17 +988,17 @@ judgment_fn! {
                 // If `'0: T` then `'0` must hold for entire fn body...
                 Parameter::Ty(_) => false,
 
-                Parameter::Lt(lt) => match lt.data() {
+                Parameter::Lt(lt) => match &**lt {
```
nikomatsakis: Hmm. I don't love this, but then... I think we should probably rewrite this to something more judgment-y...

nikomatsakis (approved):
Huh, wow, this was very smooth. I wonder if there are `&**` and things that can be removed now? Overall this seems like a net win so far, @System625 did you look at all for that?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/321/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..0b35ef3a960ed9c0afdd7871d03c35f08da0068b)*

System625:
Thanks for the review! Yes, I took a look and there was quite a bit to clean up. So I pushed a follow-up commit that:

- Removes the 6 manual `UpcastFrom`/`DowncastTo` impl blocks in `ty.rs` by using `cast_impl!` chains through `Arc<T>` (leveraging the `Arc<T> <-> T` impls from #255):
  ```rust
  cast_impl!((Ty) <: (Arc<Ty>) <: (Parameter));
  cast_impl!((Lt) <: (Arc<Lt>) <: (Parameter));
  cast_impl!((Const) <: (Arc<Const>) <: (Parameter));
  ```
  This eliminates all three `(**v).clone()` patterns in the Downcast impls.
- Simplifies `Ty::to_parameter` from `Parameter::Ty(Arc::new(self.clone()))` to `self.clone().upcast()`.
- Simplifies the `&**lt` in `nll.rs` to `lt.as_ref()`.

Net -41 lines, all 269 tests still pass.

The only remaining Arc-friction is `.copied()` in `Parameter::as_variable()` as the auto-generated `as_variable()` on the enums returns `Option<&Variable>`. Removing that would need `#[customize(constructors)]` + ~15 lines of manual methods, which felt worse than keeping `.copied()`, but happy to go either way :)

nikomatsakis:
OK, I'm going to merge this, but I want to look at the code afterwards, I still feel like there is more opportunity for cleanup. In general calls to `as_ref` or `clone` is not much better in my eyes, I want it to just say `x` not `x.clone().upcast()` :)

PR #323, opened 2026-04-09
Reject moves of borrowed places

### LINKED ISSUE #298: Reject moves of borrowed places
You cannot move a place that currently has an active loan. Currently formality detects conflicts when *writing* to a borrowed place, but does not specifically check moves.

In rustc this is error E0505: "cannot move out of `x` because it is borrowed."

This may partially work already due to the existing loan-conflict checking — the test should be tried without `#[ignore]` to see if the existing write-conflict logic happens to catch it.

**Ignored tests** (`#[ignore]` in `src/test/borrowck.rs`):

- `move_out_of_borrowed_place` — move `x` while `&x` is live

xonx4l:
## What does this PR do?
This PR treat non-Copy place uses as moves for loan checking while keeping known Copy places as reads so overlapping live loans reject moves out of borrowed places as mentioned in https://github.com/rust-lang/a-mir-formality/issues/298

e.g., closes #298

<details>

**AI disclosure.**

 * No AI tools used .

**Confidence level.**

**Questions for reviewers.** None

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -433,7 +433,8 @@ judgment_fn! {
 
         (
             (borrow_check_place_expr(env, assumptions, state, place) => (place, state))
-            (access_permitted(env, assumptions, state, Access::new(AccessKind::Read, place), places_live_on_exit) => state)
+            (access_kind_for_place_use(env, assumptions, state, place) => (access_kind, state))
+            (access_permitted(env, assumptions, state, Access::new(access_kind, place), places_live_on_exit) => state)
             // FIXME(#296): also need to track that the place has been moved from
             (prove_place_is_movable(env, assumptions, state, place) => state)
             ------------------------------------------------------------ ("place")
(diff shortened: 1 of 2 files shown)


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -745,6 +746,37 @@ judgment_fn! {
...
+            (if let TyData::RigidTy(RigidTy { name: RigidName::ScalarId(_), .. }) = place.ty.data())
+            ------------------------------------------------------------ ("scalar-copy")
+            (access_kind_for_place_use(_env, _assumptions, state, _place) => (AccessKind::Read, state))
+        )
+
+        (
+            (if let TyData::RigidTy(RigidTy { name: RigidName::Ref(RefKind::Shared), .. }) = place.ty.data())
+            ------------------------------------------------------------ ("shared-ref-copy")
+            (access_kind_for_place_use(_env, _assumptions, state, _place) => (AccessKind::Read, state))
+        )
```
nikomatsakis, line 771: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/323/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..417e2b90d85e605a2f6a1edea46e45761352f51e)*

This is a good start. We should open a tracking issue for modeling the `Copy` trait properly.

nikomatsakis (approved):

PR #325, opened 2026-04-13
formality-macros: Unpack groups until first non-trivial token sequence

dingxiangfei2009:
This is originally a fix on rust-analyzer incorrectly leaving the surrounding brackets, causing macro panics, but I figured that this is a good syntactical extension anyway.

## What does this PR do?

*No issue number yet*

<details>

**AI disclosure.**

* I used an AI tool for research, autocomplete, or in other minimal ways agent

**Confidence level.**

* I am very happy with it

**Questions for reviewers.**

</details>

### DIFF
--- a/crates/formality-macros/src/lib.rs
+++ b/crates/formality-macros/src/lib.rs
@@ -1,4 +1,4 @@
-use proc_macro::{Span, TokenStream};
+use proc_macro::TokenStream;
 use quote::quote;
 use spec::FormalitySpec;
 // use syn::DeriveInput;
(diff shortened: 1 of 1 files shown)


dingxiangfei2009:
cc @BennoLossin I hope this helps!

BennoLossin:
I tested this and my errors in rust-analyzer are still present, they look like this:
image
All of these errors are fake, running `cargo check` passes at the moment... Although this might just be a config issue on my end

nikomatsakis:
@BennoLossin I've been having similar problems and it's quite frustrating. I don't quite know when they started, things *used* to work well.

crates/formality-macros/src/lib.rs
```diff
@@ -63,10 +63,20 @@ pub fn test(args: TokenStream, input: TokenStream) -> TokenStream {
...
+        match dummy {
+            proc_macro::TokenTree::Group(ref g) => {
+                if let Some(rest) = t.next() {
+                    break (dummy, rest);
+                }
+                stream = g.stream();
+            }
+            _ => break (dummy, t.next().expect("rest")),
+        }
+    };
```
nikomatsakis, line 83: I don't really understand what this is doing :( can you explain with an example?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/325/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..64274789c67bbe3b28f0c0760a898e85765bf490)*

nikomatsakis (changes requested):

crates/formality-macros/src/lib.rs
dingxiangfei2009: Sorry for the late reply, @nikomatsakis 

I am adding a test example to demonstrate the expected macro expansion. Previously I did some `eprintln` debugging and observing the server output. Apparently with a recent `rust-analyzer` release, a pedantic `Group` was added to surround the tokens passed into the `respan!` macro.

IIRC, I saw from the print-out `( $token . $token ( .... ) )`, and I managed to peel off a layer of `Group` to get `$token . $token ( ... )`, which is actually what we want.

I am thinking of fixing it upstream, but I figured that it wouldn't hurt if we just let the macro accept it by digging behind the groups.

crates/formality-macros/src/lib.rs
```diff
@@ -104,3 +121,36 @@ pub fn respan(t: TokenStream) -> TokenStream {
...
+    use super::respan_impl;
+    use quote::quote;
+
+    #[test]
+    fn test_respan_unpacking() {
+        let input = quote! { X (a + b) };
+        let output = respan_impl(input);
+        assert_eq!(output.to_string(), "a + b");
+
+        let input_wrapped = quote! { (X (a + b)) };
```
dingxiangfei2009, line 136: I was thinking, adjusting the span would also just work if we throw more brackets at it in the consequents.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/325/893464cfaa32f7fb64fe4981804ce16846864c60..ff171a5b74635e04e44acf1971e69d8895eef9ce)*

nikomatsakis:
I'm going to close this PR because I *think* it's working around bugs in rust-analyzer which I think we can resolve in other ways. But @dingxiangfei2009 if you think this is an independently good change feel free to re-open.

PR #300, opened 2026-03-30
Convert check_coherence to a judgment function

xonx4l:
## What does this PR do?
This PR Convert check_coherence to a judgment function as discussed in  https://github.com/rust-lang/a-mir-formality/issues/294 and part of tracking issue https://github.com/rust-lang/a-mir-formality/issues/288

Moves coherence checking toward judgment_fn! so proof trees are built by the macro instead of hand-assembled ProofTrees's.

Closes #294 

## Disclosure and PR context

**AI tools used:** Yes, for research.

**Confidence level:**  Feel good about it.

**Review depth:**  reviewed closely and understand it well.

**Testing:** I ran full test suite (cargo test --all)  , cargo test coherence and cargo build .

**Questions for mentors:** None

PR #326, opened 2026-04-13
Fix core crate

BennoLossin:
## What does this PR do?

Fix two issues with the core crate:
1. #300 for some unexplained reason removed `mod core_crate` and replaced it with a manual creation of an empty core crate. It did not remove the `core_crate.rs` file, which makes the change extremely strange. See [this comment](https://github.com/rust-lang/a-mir-formality/pull/300#discussion_r3071757903) for the exact change. This PR reinstates reverts that change.
2. The core crate was `push`ed into the crates list. This of course means that it isn't included in the prefix checking of all crates in front, which means that they cannot depend on the core crate. Thus move the core crate to index 0.

<details>

**AI disclosure.**

Non AI used.

**Confidence level.**

* Pretty confident, just one question regarding the blessed test in the top commit given below.

</details>

**Questions for reviewers.**

The top commit changes the position of the core crate to index 0, which obviously changes some of the diagnostics. Some of them got some strange additions though, for example `src/test/mir_typeck.rs`:
```diff
         expect_test::expect![[r#"
-            the rule "trait implied bound" at (prove_wc.rs) failed because
-              expression evaluated to an empty collection: `decls.trait_invariants()`"#]]
+            crates/formality-rust/src/prove/prove/prove/prove_via.rs:9:1: no applicable rules for prove_via { goal: Trait1(S1), via: HasPlace(?ty_1), assumptions: {}, env: Env { variables: [?ty_1, ?ty_2, ?ty_3], bias: Soundness, pending: [], allow_pending_outlives: true } }
+
+            crates/formality-rust/src/prove/prove/prove/prove_via.rs:9:1: no applicable rules for prove_via { goal: Trait1(S1), via: HasPlace(?ty_1), assumptions: {}, env: Env { variables: [?ty_1, ?ty_2], bias: Soundness, pending: [], allow_pending_outlives: true } }
+
+            crates/formality-rust/src/prove/prove/prove/prove_via.rs:9:1: no applicable rules for prove_via { goal: Trait1(S1), via: HasPlace(?ty_1), assumptions: {}, env: Env { variables: [?ty_1], bias: Soundness, pending: [], allow_pending_outlives: true } }
+
+            crates/formality-rust/src/prove/prove/prove/prove_via.rs:9:1: no applicable rules for prove_via { goal: Trait1(S1), via: PlaceRead(?ty_1, ?ty_2), assumptions: {}, env: Env { variables: [?ty_1, ?ty_2], bias: Soundness, pending: [], allow_pending_outlives: true } }"#]]
     )
 }
```

Not sure why only the `HasPlace` and `PlaceRead` traits were reported, by symmetry it should also do it for `PlaceWrite` etc. The reason that we only see these might be that the diagnostics are limited to 4, but I haven't checked that.


crates/formality-rust/src/check/mod.rs
```diff
@@ -34,22 +35,18 @@ judgment_fn! {
...
-            } else {
+            (let crates = {
                 let Crates { mut crates } = crates.clone();
-                // Empty core when the program omits it.
-                crates.push(Crate {
-                    id: CrateId::new("core"),
-                    items: vec![],
-                });
-                Crates { crates: crates }
+                if Some("core") != crates.first().map(|first| &**first.id) {
```
nikomatsakis, line 40: were we going to rename this to "lang-items"?

nikomatsakis (approved):
OK, this seems good. There was some confusion around the transition to judgments, I'm not surprised some logic got lost in all the rebasing and so on.

I'll land it but also @BennoLossin maybe we want to rename things?

As for the diagnostics, I know why they are occurring, we should probably improve them a bit but it's ok. It's I think because you're bringing new traits which is creating new opportunities for implied bounds.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/326/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..8166b6a9ca50aded41ed821e564b7e77c6ba5de9)*

PR #327, opened 2026-04-13
Rename field projections lang items to match rustc

BennoLossin:
## What does this PR do?

Rename the lang items for field projections to match rustc.

While writing some docs for std, I came up with some better names & then discussed this in the custom-refs zulip channel. We should have the same names here to avoid confusion.

<details>

**AI disclosure.**

**Confidence level.**

* I am very happy with it

**Questions for reviewers.**

</details>

This depends on #326 and is stacked on top, since it needs to change the names also in the diagnostic output. That PR should be merged first.


lqd (approved):
I only looked at the commit that's not in the previous PR: looks good. This should be ready to land after the other lands

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/327/288f923ceac3e1c6e3eb382a01f8a2c2c29e5168..90bab761ed1be0a5139a714f4b51c3e7cde462f9)*

Issue #304, opened 2026-04-03
Investigate why struct test is failing in WF check


nikomatsakis:
PR #282 left an odd error in `src/test/borrowck.rs`. See the conversation here:

https://github.com/rust-lang/a-mir-formality/pull/282#discussion_r3019048262

PR #331, opened 2026-04-15
grammar: resolve PlaceExpr ambiguity with precedence levels

pranavpurankar:
## What does this PR do?

Closes #334 

This PR attempts to resolve the `*p.f` ambiguity. After testing, I've found that the `#[precedence]` attribute does not behave as expected for mixed prefix (`*`) and infix (`.`) operators. Resulting -> The parser currently cannot distinguish between *(w.value) and  (*w).value

### Background Research
I reviewed the parser implementation in `formality-core`:
* [parser.rs](https://github.com/rust-lang/a-mir-formality/blob/56e5c2a9b41375cf7c89a4b8d3605b662895ecec/crates/formality-core/src/parse/parser.rs#L46-L51) / [precedence.rs](https://github.com/rust-lang/a-mir-formality/blob/main/crates/formality-macros/src/precedence.rs#L63): Precedence is designed so higher numbers = tighter binding.
* [left_associative.rs](https://github.com/rust-lang/a-mir-formality/blob/56e5c2a9b41375cf7c89a4b8d3605b662895ecec/tests/parser-torture-tests/left_associative.rs#L108): This works perfectly for **infix binary operators** (e.g., `Add(1)` vs `Mul(2)`).

_I've tried to understand above code blocks mentioned in above files and this files (most of it went right above my head )_, also checked [grammar.rs](https://github.com/rust-lang/a-mir-formality/blob/56e5c2a9b41375cf7c89a4b8d3605b662895ecec/examples/formality-eg/grammar.rs#L79), [right_associative](https://github.com/rust-lang/a-mir-formality/blob/56e5c2a9b41375cf7c89a4b8d3605b662895ecec/tests/parser-torture-tests/right_associative.rs#L84)

However this issue is counterintuitive to me because of simple thought PEMDAS & predence,  associativity rules, so mixing a prefix operator (`Deref`) and an infix operator (`Field`) across separate variants, the parser fails.

### Experimental Configurations
I tested the following configurations.

| Config | Deref Priority | Field Priority | Result | Analysis |
| :--- | :--- | :--- | :--- | :--- |
| **A (Inverted)** | 10 | 1 | ✅ Parses (No Panic) | **Semantically Wrong**: Produces `Field(Deref(p), f)` instead of `Deref(Field(p, f))`. Proves the attribute is read, but forces the wrong tree. |
| **B (Standard)** | 1 | 2 | ❌ Ambiguity Panic | Identifies both interpretations but refuses to choose, despite Field having higher precedence. |
[...]

PR #309, opened 2026-04-06
Fix well-formedness assumptions

xonx4l:
## What does this PR do?
This PR  fix well-formedness assumptions and  address issue https://github.com/rust-lang/a-mir-formality/issues/304

Closes #304

## Disclosure and PR context

**AI tools used:** Yes , For research .

**Confidence level:** initial draft I need to work more on some specifics  

**Review depth:** I reviewed closely and understand it well

**Testing:** cargo test , cargo fmt check

**Questions for mentors:** None

PR #332, opened 2026-04-15
Fix: well-formedness assumptions for existential variables

pranavpurankar:
## What does this PR do?

closes PR #304 

This PR resolves the well-formedness (WF) check failure in the NLL borrow checker that was identified during the work on #282.
Previously, it had no assumptions regarding WF of these variables (_fresh existential lifetime_), causing prove_wf to fail with "no applicable rules" before even it hit borrow checker

### Traceability

0. Zulip discussion about test and wf assumption inception -> [#t-types/formality > &#91;GFI&#93; Expand test coverage for Expr variants formality#265 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/.5BGFI.5D.20Expand.20test.20coverage.20for.20Expr.20variants.20formality.23265/near/582856965)

1. In the PR #282 -> test `struct_with_mutable_reference_locks_local` encountered an ambigious parse Err [[Code block](https://github.com/rust-lang/a-mir-formality/pull/282/changes#diff-8808332714b76180e9f23145b85aa243adeccc4c987378d2c13342afe6be5b40R1293)]

2. PlaceExpr ambiguity precedence levels also occured in the same test, PR #331 resolved it.

3. Possibility of ripple effect in the attempt to solve WF #309 

<details>

**AI disclosure.**

- I used AI tool for mapping DSL syntax, .fold() and upcast() understanding and conciseness of code (verbosity of code reduction)

**Confidence level.**

* I am very happy with it, as I followed suggestion provided by @nikomatsakis and it took decent amount of time to understand exact fix and problem.

**Questions for reviewers.**
@nikomatsakis  Does it meet the direction that you were trying to provide here #309?
An err appearing right now is expected one? Because I didn't understood it correctly as compared to other errors in the codebase (at least basic intepretation of an err). 

</details>

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -281,6 +281,13 @@ judgment_fn! {
         (
             // Existential variables: open the binder and check the inner block
             (let (env, subst, block) = env.instantiate_existentially(binder))
+            (let wf_assumptions: Wcs = subst.iter().fold(Wcs::default(), |acc, v| {
+                let p: Parameter = v.upcast();
+                let r: Relation = p.well_formed();
+                let w: Wcs = r.upcast();
+                (acc, w).upcast()
+            }))
+            (let assumptions: Wcs = (assumptions, wf_assumptions).upcast())
             (borrow_check_block(env, assumptions, state, block, places_live_on_exit) => state)
             (let state = state.pop_subst(&env.env, subst))
             ------------------------------------------------------------ ("exists")
--- a/src/test/borrowck.rs
+++ b/src/test/borrowck.rs
@@ -931,7 +931,10 @@ fn undeclared_universal_region_relationship() {
             }
         ]
 
-        expect_test::expect!["crates/formality-rust/src/prove/prove/prove/prove_outlives.rs:8:1: no applicable rules for prove_outlives { a: !lt_0, b: !lt_1, assumptions: {}, env: Env { variables: [!lt_0, !lt_1], bias: Soundness, pending: [], allow_pending_outlives: false } }"]
+        expect_test::expect![[r#"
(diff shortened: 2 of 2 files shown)


nikomatsakis:
I *think* this overlaps with #309 -- I'll compare

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -281,6 +281,13 @@ judgment_fn! {
         (
             // Existential variables: open the binder and check the inner block
             (let (env, subst, block) = env.instantiate_existentially(binder))
+            (let wf_assumptions: Wcs = subst.iter().fold(Wcs::default(), |acc, v| {
+                let p: Parameter = v.upcast();
+                let r: Relation = p.well_formed();
+                let w: Wcs = r.upcast();
+                (acc, w).upcast()
+            }))
```
nikomatsakis, line 289: OK, so, I agree this is kind of the "minimal PR", but I think I prefer the approach in #309 where we add a new `Upcast` impl for `Wcs` and use it to cleanup this code (and other code). Let me check in on that PR.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/332/ef43ba03243c2dc83db2f5cd4b85e366716b906f..e358ff2591e7e3b6ef5cb4dc6af280ac76dc1b91)*

pranavpurankar: Alright, so I was missing the big picture -> making it more robust at foundational level and isloated specific logic i.e blocker. I will keep it open until #309 lands or need to think about it

nikomatsakis:
Gonna close this one in favor of #309 and #340. Thanks @pranavpurankar !

PR #342, opened 2026-04-21
make deref place exprs use the `Derefable` trait

BennoLossin:
## What does this PR do?

This is a WIP change to make the deref place expression use the `Place` trait instead of expecting a reference. This is the first step to adding virtual places.

@nikomatsakis I'd like to get your opinion on how to proceed with this. I'm a bit stuck because the judgement `fn borrow_check_place_expr` does not return a constraint, so I cannot bubble up any additional constraints created by `prove`. What would be the best approach to solve that?

I don't know if I already need to adjust `fn prove_place_is_movable`, `fn access_permitted` and more, they seem like they could stay as-is for now and only need adjustments when I allow custom reading, writing & borrowing. But I'm a bit unsure...

<details>

**AI disclosure.**

* None used

**Confidence level.**

* Early draft, want feedback

**Questions for reviewers.**

</details>


BennoLossin:
I think this is ready for review.

will try to add the traits for the other place expressions & the place ops with borrowck knobs incrementally

nikomatsakis (approved):

PR #344, opened 2026-04-23
add rust-analyzer to toolchain file and choose working nightly version

BennoLossin:
fixes the errors rust-analyzer was displaying for judgement functions on my side. Would appreciate some testing from other people especially non-nix users.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2025-12-01"
-components = ["rustfmt"]
+channel = "nightly-2025-11-24"
+components = ["rustfmt", "rust-analyzer"]


dingxiangfei2009 (approved):
Yeah, this version worked for me, too.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/344/893464cfaa32f7fb64fe4981804ce16846864c60..33c15a83535709b01ddce7e5eed6acb10068af00)*

nikomatsakis (approved):

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2025-12-01"
-components = ["rustfmt"]
+channel = "nightly-2025-11-24"
```
nikomatsakis, line 2: Wait. Why did you change this *back*? That date comes from mini-rust so I don't want to change it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/344/893464cfaa32f7fb64fe4981804ce16846864c60..33c15a83535709b01ddce7e5eed6acb10068af00)*

BennoLossin: I ran a bisect on the r-a version and 2025-11-25 is the earliest that's failing. Some more details on zulip: https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/rust-analyzer.20errors/near/590543291

nikomatsakis: Hmm. So basically there's some bug in rust-analyzer we are working around?

Veykril: This likely stems from https://github.com/rust-lang/rust-analyzer/pull/21097 and some encoding mismatch between the client and server versions. At this point I think we consider this acceptable breakage as the toolchain version you are using is so much older than current rust-analyzer. As for why 2025-12-01 is specifically not working, there were a bunch of bugs after this PR that we slowly fixed, so it might very well be that this nightly range is just overall broken wrt to proc-macros

BennoLossin: I think the reason for pinning it to the December version is mini-rust (or a dependency)? So also including rust-analyzer in that should be correct, right?

So @nikomatsakis we can maybe "just" merge this and then later increment the pinned version to one that works (since mini-rust or the dependency was incremented by Ralf IIRC)?

Veykril: Yea, including rust-analyzer there will make VSCode pick up the corresponding rust-analyzer instead of using the bundled one. (zed iirc does not do that yet)

Ideally you'd do that AND update mini-rust and friends so you can bump the toolchain file to a nightly where r-a isn't as buggy

BennoLossin:
I haven't followed the repo updates, do we still need this PR?

BennoLossin:
I just checked this again, I still get errors without this patch, so I'll leave this open. I think regardless of which rust version we end up pinning, rust-analyzer should also be pinned, since then it won't stop working in the future.

If we don't end up pinning the Rust version, then I think we can close this.

Veykril:
> I think regardless of which rust version we end up pinning, rust-analyzer should also be pinned, since then it won't stop working in the future.

^ generally if you pin your toolchain you want to add the rust-analyzer component to the toolchain file

https://github.com/rust-lang/a-mir-formality/pull/369 fixes the r-a errors for you fwiw, the current pinned toolchain just has a buggy r-a

nikomatsakis:
Closing in favor of #369

PR #346, opened 2026-04-24
Use boolean literals instead of digit

liolin:
## What does this PR do?

Replaces `0 _ bool` with `false` in test cases. This simplifies the conversion to Rust as no special case is needed.

<details>

**AI disclosure.**

* No AI used.

**Confidence level.**

* I am very happy with it

**Questions for reviewers.**

Was there a specific reason to use `0 _ bool`?

</details>

### DIFF
--- a/crates/formality-rust/src/grammar/expr/mod.rs
+++ b/crates/formality-rust/src/grammar/expr/mod.rs
@@ -132,6 +132,7 @@ pub enum ExprData {
     ///
     /// A scalar literal.
     #[grammar($value _ $ty)]
+    #[reject(_, ScalarId::Bool)]
     Literal { value: usize, ty: ScalarId },
 
     /// `true`
--- a/src/test/mir_typeck.rs
+++ b/src/test/mir_typeck.rs
@@ -30,7 +30,7 @@ fn test_assign_constant() {
                     let v7: i32 = 5 _ i32;
                     let v8: i64 = 5 _ i64;
                     let v9: isize = 5 _ isize;
-                    let v10: bool = 0 _ bool;
+                    let v10: bool = false;
                     return 5 _ u8;
                 }
             }
@@ -156,7 +156,7 @@ fn if_else_different_return_types() {
                     if b {
                         return 1 _ u32;
                     } else {
-                        return 0 _ bool;
+                        return false;
                     }
                 }
             }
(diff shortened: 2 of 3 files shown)


lqd (approved):

src/test/mir_typeck.rs
```diff
@@ -30,7 +30,7 @@ fn test_assign_constant() {
                     let v7: i32 = 5 _ i32;
                     let v8: i64 = 5 _ i64;
                     let v9: isize = 5 _ isize;
-                    let v10: bool = 0 _ bool;
+                    let v10: bool = false;
```
nikomatsakis, line 33: Hmm, I think we should *probably* make `0_bool` an error while we are at it. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/346/893464cfaa32f7fb64fe4981804ce16846864c60..b23d45714af94aef8222fc2937377f7aa16dbcdd)*

nikomatsakis (changes requested):

crates/formality-rust/src/grammar/expr/mod.rs
```diff
@@ -132,6 +132,7 @@ pub enum ExprData {
     ///
     /// A scalar literal.
     #[grammar($value _ $ty)]
+    #[reject(_, ScalarId::Bool)]
```
lqd, line 135: can you also add a test that exercises the failure?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/346/893464cfaa32f7fb64fe4981804ce16846864c60..e82908fa51079406c1fcf84be5c6d1da53bab011)*

liolin: Where would you like to have the test (in which file)?

lqd: it doesn't matter that much, I guess close to where we test parsing makes sense?

lqd (approved):
Great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/346/328048a2411c7414ac96a06593b51b70f91ee9b5..9ad9fc6eb2fbc8787b9f3359ac2e0b4ea031c9a4)*

nikomatsakis (approved):

PR #347, opened 2026-04-25
Lower the lifetimes introduced by an `exists` statement to Rust

liolin:
## What does this PR do?

Lowers an `exists` statement. The newly introduced lifetimes e.g. `'r0`, are all treated as anonymous lifetimes `'_`.

```rust,ignore
exists<'r0> {
  let v2: &'r0 u32 = v1;
  return v2;
}
```
becomes:
```rust,ignore
{
  let v2: &'_ u32 = v1;
  return v2;
}
```
<details>

**AI disclosure.**

* No AI used.

**Confidence level.**

**Questions for reviewers.**

* Is using `'_` for an existential variable the right choice?

</details>

### DIFF
--- a/crates/formality-rust/src/to_rust/expr.rs
+++ b/crates/formality-rust/src/to_rust/expr.rs
@@ -70,10 +70,29 @@ impl RustBuilder {
         })
     }
 
+    /// Lowers an `exists` statement. The newly introduced lifetimes
+    /// e.g. `'r0`, are all treated as anonymous lifetimes `'_`.
+    /// ```rust,ignore
+    /// exists<'r0> {
+    ///   let v2: &'r0 u32 = v1;
+    ///   return v2;
+    /// }
+    /// ```
+    /// becomes:
+    /// ```rust,ignore
+    /// {
+    ///   let v2: &'_ u32 = v1;
+    ///   return v2;
+    /// }
+    /// ```
     fn lower_exists_stmt(&mut self, binder: &Binder<Block>) -> Fallible<syntax::Stmt> {
-        self.with_binder(binder, |term, pp| {
-            Ok(syntax::Stmt::Block(pp.lower_block(term)?))
-        })
+        let term = binder.peek();
+        self.ctx.push_anonymous(binder.kinds());
+
+        let result = syntax::Stmt::Block(self.lower_block(term)?);
+
+        self.ctx.pop();
+        Ok(result)
     }
 
     pub fn lower_expr(&mut self, expr: &Expr) -> Fallible<syntax::Expr> {
--- a/crates/formality-rust/src/to_rust/expr.rs
+++ b/crates/formality-rust/src/to_rust/expr.rs
@@ -62,7 +62,7 @@ impl RustBuilder {
         init: Option<&Init>,
     ) -> Fallible<syntax::Stmt> {
         Ok(syntax::Stmt::Let {
-            // TODO: is there a way to knwo, if the variable must be mutable or not?
+            // TODO: is there a way to know, if the variable must be mutable or not?
             mutable: true,
             name: id.deref().clone(),
             ty: self.lower_ty(ty)?,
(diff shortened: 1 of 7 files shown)


lqd:
You shouldn’t even need to use the anonymous lifetime, and elide them entirely.

liolin:
I've updated my code. I assume the `exists` statement creates existential variables. However, this is not what the current parser creates. I must first clarify what the expected behavior is ([See Zulip](https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/.60exists.60.20introduces.20bound.20variables/with/590973508))

liolin:
I've updated the code. Existential lifetimes are now elided. When opening an `exists` binder, the variables are replaced with new existential variables. To have the same / similar procedure for bound, existential and universal variables, I refactored a lot of the "stringification" of the `Variable` type.

crates/formality-rust/src/to_rust/expr.rs
```diff
@@ -229,4 +233,30 @@ fn foo() -> () {
...
+                            return *v2;
+                        }
+                    }
+                }
+            ],
+            r#"
+fn foo() -> u32 {
+    {
+        let mut v1: u32 = 0_u32;
+        let mut v2: &mut u32 = &mut v1;
```
nikomatsakis, line 256: FYI, in my codegen branch, I've introduced an "Erased" lifetime, it hasn't merged yet, but it might be useful here too

crates/formality-rust/src/to_rust/mod.rs
```diff
@@ -76,122 +74,120 @@ pub fn create_workspace(crates: &Crates, root_directory: &std::path::Path) -> Fa
...
+}
 
-    pub fn current_names(&self) -> Fallible<&[String]> {
-        self.variable_names
-            .first()
-            .map(|v| v.as_slice())
-            .ok_or_else(|| anyhow::anyhow!("no active binder frame"))
-    }
+impl RustBuilder {
+    pub fn with_binder<T: Fold, R>(
```
nikomatsakis, line 125: can you add a comment explaining what is going on here? the ideal is to include some example rust snippet

crates/formality-rust/src/to_rust/mod.rs
```diff
@@ -76,122 +74,120 @@ pub fn create_workspace(crates: &Crates, root_directory: &std::path::Path) -> Fa
...
+        &mut self,
+        b: &Binder<T>,
+        is_trait: bool,
+        g: impl Fn(&T) -> &Vec<WhereClause>,
```
nikomatsakis:
```suggestion
        g: impl Fn(&T) -> &[WhereClause],
```

if you use `&[]` then we can return constants if that ever becomes relevant, e.g., an empty slice

crates/formality-rust/src/to_rust/mod.rs
```diff
@@ -76,122 +74,120 @@ pub fn create_workspace(crates: &Crates, root_directory: &std::path::Path) -> Fa
...
-        self.variable_names
-            .first()
-            .map(|v| v.as_slice())
-            .ok_or_else(|| anyhow::anyhow!("no active binder frame"))
-    }
+impl RustBuilder {
+    pub fn with_binder<T: Fold, R>(
+        &mut self,
+        b: &Binder<T>,
+        is_trait: bool,
```
nikomatsakis, line 128: be sure to comment what's going on here :)

nikomatsakis (changes requested):
This looks close!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/347/0767b6686cf6c854d6166ed0285dceafdd2dffa4..f0b70e38b667318eb9e2e924d1ed950ac5db1e26)*

crates/formality-rust/src/to_rust/mod.rs
liolin: I've added doc comments.

liolin:
After this PR is merged, I would like to rename the `RustBuilder` struct to something like `Context` and make all the functions freestanding functions. I think that would make certain things more elegant.

nikomatsakis (approved):

nikomatsakis (approved):

PR #349, opened 2026-04-28
Struct lt wf

BennoLossin:
(depends on #344)

## What does this PR do?

@_**Benno Lossin|458499** [said](https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/Office.20hours/near/590222965):
> @**nikomatsakis** I tried to make it work, but the following test already doesn't work:

```rust
#[test]
fn missing_constraint() {
    crate::assert_err!(
        [
            crate test {
                struct Foo<'a> {
                    y: &'a u32,
                }

                fn foo() -> () {
                    exists<'y> {
                        let a: u32 = 22 _ u32;
                        let f: Foo<'y> = Foo { y: &'y a };
                    }
                }
            }
        ]
        expect_test::expect![[r#"
            crates/formality-rust/src/prove/prove/prove/prove_via.rs:9:1: no applicable rules for prove_via { goal: @ wf(Foo<?lt_0>), via: @ wf(?lt_0), assumptions: {@ wf(?lt_0)}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }

            crates/formality-rust/src/prove/prove/prove/prove_wf.rs:14:1: no applicable rules for prove_wf { goal: ?lt_0, assumptions: {@ wf(?lt_0)}, env: Env { variables: [?lt_0], bias: Soundness, pending: [], allow_pending_outlives: true } }"#]]
    )
}
```

> This seems wrong?

Xiang and I looked into this a bit in our meeting today and we think that we actually found the issue:

`prove_wf` doesn't have a rule to prove wf via assumption and it is directly called from itself (when checking wf of parameters for example), so any assumption proving logic in `prove_wc` doesn't get used.

There are two solutions:
- call `prove_wc` or `prove` from `prove_wf` instead of directly recursing,
- add a `via-assumption` rule to `prove_wf`

I have implemented the latter one (since I thought it would be less overall churn) to see what other things break/change. There are some test output changes and I'll send a PR.

From an overall design perspective, I think we should either handle assumptions *once* at the very top in `prove`. Or each `prove_*` function should do it themselves.

<details>

**AI disclosure.**

No ai used.

**Confidence level.**

</details>

### DIFF
--- a/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
+++ b/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
@@ -1,6 +1,9 @@
-use crate::grammar::{
-    AliasName, AliasTy, ConstData, Lt, LtData, Parameter, Parameters, Relation, RigidName, RigidTy,
-    Ty, UniversalVar, Wcs,
+use crate::{
+    grammar::{
+        AliasName, AliasTy, ConstData, Lt, LtData, Parameter, Parameters, Relation, RigidName,
+        RigidTy, Ty, UniversalVar, Wcs,
+    },
+    prove::prove::prove::prove_via::prove_via,
 };
 use formality_core::{judgment_fn, Downcast, ProvenSet};
 
@@ -21,6 +24,13 @@ judgment_fn! {
         debug(goal, assumptions, env)
 
         assert(env.encloses((assumptions, goal)))
+
+        (
+            (a in assumptions)!
+            (prove_via(decls, env, assumptions, a, Relation::WellFormed(goal.clone())) => c)
+            --- ("by-assumption")
+            (prove_wf(decls, env, assumptions, goal) => c)
+        )
 
         (
             // Always assume that universal variables are WF. This is debatable, it implies
--- a/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
+++ b/crates/formality-rust/src/prove/prove/prove/prove_wf.rs
@@ -2,11 +2,11 @@ use crate::grammar::{
     AliasName, AliasTy, ConstData, Lt, LtData, Parameter, Parameters, Relation, RigidName, RigidTy,
     Ty, UniversalVar, Wcs,
 };
-use formality_core::{judgment_fn, Downcast, ProvenSet};
+use formality_core::{judgment_fn, Downcast, ProvenSet, Upcast};
 
 use crate::prove::prove::{
     decls::Program,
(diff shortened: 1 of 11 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2025-12-01"
-components = ["rustfmt"]
+channel = "nightly-2025-11-24"
```
nikomatsakis: let's not do this and/or put it in a separate PR

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/349/cc48e5d27171ad19bbb0b99653f43de7971cb026..6023fca58785650e28fb22305533cd4d633495b4)*

BennoLossin:
I totally forgot about this PR...

@nikomatsakis I updated the implementation as well, instead of having an `by-assumption` rule in `prove_wf`, instead it now recurses into `prove`, which handles assumptions.

crates/formality-rust/src/prove/prove/prove/prove_wf.rs
```diff
@@ -95,5 +95,14 @@ pub fn prove_alias_wf(
...
+    for_all(decls, env, assumptions, parameters, &prove_wf_recursive)
+}
+
+pub fn prove_wf_recursive(
+    program: impl Upcast<Program>,
+    env: impl Upcast<Env>,
+    assumptions: impl Upcast<Wcs>,
+    param: impl Upcast<Parameter>,
+) -> ProvenSet<Constraints> {
+    prove(program, env, assumptions, Relation::well_formed(param))
```
nikomatsakis, line 107: I'd probably just inline this, but I guess this is fine

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/349/cc48e5d27171ad19bbb0b99653f43de7971cb026..6023fca58785650e28fb22305533cd4d633495b4)*

nikomatsakis (approved):

crates/formality-rust/src/prove/prove/prove/prove_wf.rs
BennoLossin: I call the function 6 times, 4 of which are in `for_all`, which doesn't give parameters to `prove_wf_recursive`. Having a closure there would be pretty unreadable IMO

PR #338, opened 2026-04-15
Flatten Wc/WcData into a single enum

System625:
## What does this PR do?

Opened PR 2 of the #318 series, it flattens Wc/WcData using the same pattern as #321. Tried to split it smaller but the cast chains don't compose across a seam, so it's wc.rs plus the direct consumers that wouldn't compile otherwise. All tests green, no snapshot regen needed

<details>
<summary>Disclosure questions</summary>

**AI disclosure.**

* The AI tool authored large parts of the code

**Confidence level.**

* It seems reasonable to me, I'd like to know what others think

**Questions for reviewers.**
None

</details>

PR #354, opened 2026-05-05
Flatten PlaceExpr/PlaceExprData into a single enum

System625:
## What does this PR do?
Same pattern as #338 (Wc) and #353 (WhereClause): drop the wrapper struct, promote the enum to be the type, keep PlaceExprData as a migration alias. Recursive variants hold Arc<PlaceExpr> directly.

Part of #318 

After the flatten, the recursion is still broken by Arc, just placed on the recursive fields themselves. So sharing/cheap-clone behavior is preserved.
I kept pub type PlaceExprData = PlaceExpr as a migration alias so any code (or future PRs) that still says PlaceExprData::Var(..) continues to work. The methods (prefix, all_prefixes, is_prefix_of) moved straight onto the enum and now match on self directly instead of going through self.data().
Same pattern as #338 (Wc) and #353 (WhereClause), so I'm following an established shape and that's part of why I'm fairly confident it's correct. I will wait for this to be reviewed and merged before I start working on the flattening of Expr/ExprData which has a bigger blast radius than this current one.
Quick question:
 - Is the pub type PlaceExprData = PlaceExpr alias still wanted, or should I just rename all call sites in
  this PR? Since this leaves only Arc<T> as the remaining wrapper in the grammar, which I think is what unblocks the
  precedence-bug fix discussed in office hours.

## AI disclosure

### DIFF
--- a/crates/formality-rust/src/grammar/expr/mod.rs
+++ b/crates/formality-rust/src/grammar/expr/mod.rs
@@ -212,8 +212,6 @@ pub enum PlaceExpr {
 }
 // ANCHOR_END: PlaceExpr
 
-pub type PlaceExprData = PlaceExpr;
-
 impl PlaceExpr {
     pub fn is_prefix_of(&self, other: &PlaceExpr) -> bool {
         other.all_prefixes().contains(&self)
--- a/crates/formality-rust/src/grammar/expr/mod.rs
+++ b/crates/formality-rust/src/grammar/expr/mod.rs
@@ -2,7 +2,7 @@ mod codegen;
 
 use std::sync::Arc;
 
-use formality_core::{cast_impl, id, term, DowncastTo, UpcastFrom};
+use formality_core::{id, term, DowncastTo};
 
 use crate::grammar::{
     AdtId, Binder, FieldName, Lt, Parameter, RefKind, ScalarId, TraitId, Ty, ValueId,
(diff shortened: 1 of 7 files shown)


nikomatsakis:
> I kept pub type PlaceExprData = PlaceExpr as a migration alias so any code (or future PRs) that still says PlaceExprData::Var(..) continues to work.

I wouldn't bother with this. I'd rather just fix that code.

crates/formality-rust/src/grammar/expr/mod.rs
```diff
@@ -242,27 +191,55 @@ pub enum PlaceExprData {
...
+        prefix: Arc<PlaceExpr>,
         field_name: FieldName,
     },
 }
-// ANCHOR_END: PlaceExprData
+// ANCHOR_END: PlaceExpr
 
-cast_impl!((ValueId) <: (PlaceExprData) <: (PlaceExpr));
+pub type PlaceExprData = PlaceExpr;
+
```
nikomatsakis: Not worth it! Remove these.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/354/328048a2411c7414ac96a06593b51b70f91ee9b5..aa0a73004e72ee08a49a06e009f926a7d01b2524)*

nikomatsakis (changes requested):

nikomatsakis (approved):

PR #355, opened 2026-05-08
Refactor `to_rust` module

liolin:
## What does this PR do?

Refactors the `to_rust` module to use freestanding functions. The handling of binders is unified and simplified.

Transforming methods to freestanding functions and renaming `RustBuilder` to `Context`

The previous `with_binder` function had a complicated type signature and didn't feel right. The replacement function `open_bounded` instantiates the binder with a new substitution and returns the term including the newly generated variable names.

To prevent replication (opening binder followed by generating the generics), I implemented the macro `open_bounded!` that calls the `open_bounded` function on Context and generates the Generics afterward.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/to_rust/context.rs
+++ b/crates/formality-rust/src/to_rust/context.rs
@@ -191,4 +191,4 @@ macro_rules! open_bounded {
         (term, generics)
     }};
 }
-pub(crate) use open_bounded; // <-- the trick
+pub(crate) use open_bounded;
--- a/crates/formality-rust/src/to_rust/context.rs
+++ b/crates/formality-rust/src/to_rust/context.rs
@@ -1,0 +1,182 @@
+use crate::{
+    grammar::{
+        Binder, BoundVar, ExistentialVar, Fallible, ParameterKind, UniversalVar, VarIndex,
+        Variable, WhereClause,
+    },
+    rust::Fold,
+    to_rust::{lower_generics_for_binder, syntax},
+};
+
+#[derive(Debug)]
+pub struct Context {
+    counter: VarIndex,
+}
+
+impl Context {
+    pub fn empty() -> Context {
+        Context::default()
+    }
+}
+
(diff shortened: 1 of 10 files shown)


lqd:
Generally looks good to me, though I think `Context` may be too generic of a name for one leg of the multiple code generation legs planned: if both generating-to-rust and generating-to-minirust lowering procedures need contexts, then `Context` will not refer to either. (Fine if it's shared by both, not great if it's not)

liolin:
What do you think would be a better name? Something like `ToRustContext`?

lqd:
The WIP MiniRust codegen in #352 calls it `CodegenGlobal` I guess, which is also suboptimal. `RustBuilder` and `MiniRustBuilder` seemed fine I dunno. I'll defer to @nikomatsakis.

nikomatsakis:
@liolin can you clarify for me, what is the reason to move towards free functions? I don't object but I think I missed something,.

liolin:
I do not like how `RustBuilder` has a lot of functions attached. Especially, because a lot of them don't alter the state, but only pass it to the next function. The other problem I have with that solution is that the docs (`cargo doc`) for the `RustBuilder` is huge.

However, that is only my personal preference (maybe I'm influence by Haskell), and happy to change it back if this is your preferred style.

nikomatsakis:
I'm ok either way, I just wondered if there was something else I was missing.

liolin:
@nikomatsakis I rebased my PR. Should I change the name of the `Context` to something different?

nikomatsakis (approved):
I think it's ok for now, easy enough to rename later

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/355/ad33bf5f099bb673b85b48ce3e71169c505cf220..c0b1a035e09d0d9e3ca77873b0cba22b93e7708c)*

PR #357, opened 2026-05-10
Consolidate test files

### LINKED ISSUE #308: Consolidate the test files
Currently, our test is in both ``src/test`` and ``tests``, it will be nice to have only one place for all tests.

tiif:
## What does this PR do?

Move all tests from ``src/test`` to ``tests``. 

Previously when we added test in ``src/test/``, we always need to add a ``mod test_name`` in ``src/test/mod.rs``. After moving the tests to ``tests/`` we no longer need to do that, yay!

Closes #308

## AI disclosure

* I did not use any AI tools


tests/coherence_overlap.rs
tiif, line 1: We had both src/test/coherence_overlap and tests/coherence_overlap, so the diff is verbose.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/357/6a56d765210d5611eab6a6487996fc22b0318fbb..914794c6d9c66328b6d59218e0df1b0e23e59d35)*

lqd:
Looks good, using only `tests` will prevent from having internal only testing stuff, but for such high level tests we've been doing and that this PR moves, integration tests like these are the right move IMO.

If we need to test internal procedures in the future, we can add some `src/` tests then.

lqd (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/357/6a56d765210d5611eab6a6487996fc22b0318fbb..914794c6d9c66328b6d59218e0df1b0e23e59d35)*

PR #361, opened 2026-05-14
Add positive-test coverage tracking

System625:
## What does this PR do?
Adds positive-test coverage tracking. Every time a positive test proves a judgment, we walk the `ProofTree` and append a JSONL line to `target/test-coverage.jsonl` recording the test name and every `(judgment, rule)` pair that fired. First step toward the report/front-end described in the project notes.

- New module `formality_core::judgment::coverage` (generic, language-independent).
- Hooked into `ProvenSet::assert_ok` and the `assert_ok!` macro.
- Integration test under `crates/formality-core/tests/coverage.rs` exercises the recorder with a tiny dummy language.
- Opt-out via `FORMALITY_COVERAGE=0`; output dir overridable via `FORMALITY_COVERAGE_DIR`.

The proof tree comes from positive tests as they end up calling either `ProvenSet::assert_ok` or `test_program_ok`, both of which already produce/hold a `ProofTree`, so we just read it. `ProvenSet::assert_ok` is the generic hook and covers most positive tests including `test_where_clause`. The `assert_ok!` macro is the second hook, since it goes through `test_program_ok` which returns a bare `ProofTree` and bypasses `ProvenSet`. JSONL is used because `cargo test` runs in parallel as small append-only writes are atomic, so multiple test processes can append without locking. No questions

## AI disclosure

### DIFF
--- a/crates/formality-core/src/judgment/coverage.rs
+++ b/crates/formality-core/src/judgment/coverage.rs
@@ -1,0 +1,167 @@
+//! Coverage tracking for positive judgment tests.
+//!
+//! When a test successfully proves a judgment, we walk its [`ProofTree`] and
+//! record every `(judgment, rule)` pair that fired. Records are appended,
+//! one JSONL line per test, to `target/test-coverage.jsonl`. The eventual goal
+//! is to identify judgment rules that are not exercised by any positive test.
+//!
+//! Recording is best-effort: I/O errors are swallowed so coverage never fails
+//! a test. Disable entirely with `FORMALITY_COVERAGE=0`. Redirect the output
+//! directory with `FORMALITY_COVERAGE_DIR` (used by the integration tests).
+
+use std::collections::BTreeSet;
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::panic::Location;
+use std::path::PathBuf;
+
+use super::ProofTree;
+
+/// A single (judgment, rule) pair as identified in the coverage output.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+pub struct RuleId {
+    pub judgment: String,
+    pub rule: String,
+    pub file: String,
+    pub line: u32,
+}
+
+impl ProofTree {
+    /// Walk this proof tree and return every distinct `(judgment, rule)` pair
+    /// that fired. Nodes with no `rule_name` (e.g. `for_all`, `if`, `let` leaves)
+    /// are skipped because they do not correspond to user-defined inference rules.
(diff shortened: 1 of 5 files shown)


crates/formality-core/src/judgment/coverage.rs
```diff
@@ -0,0 +1,163 @@
...
+pub fn record_coverage<'a>(trees: impl IntoIterator<Item = &'a ProofTree>) {
+    if !coverage_enabled() {
+        return;
+    }
+
+    let Some(path) = coverage_file_path() else {
+        return;
+    };
+
+    let test_name = std::thread::current()
```
nikomatsakis: "Hmm". This doesn't make sense to me. Does every test run in its own thread? I think what I would expect is to make `record_coverage` be `#[track_caller]` or else to have it take a `Location` argument and then have the caller create it. That way, we're recording the test based on the `#[test]` code. (You may need to make the *caller* `#[track_caller]` as well to get up to the right place...)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/361/73f406c98500e0382d567f72e037241b0bc5cc49..00684a850cd55baedf13b3c8abe4bcad1dbaf6db)*

nikomatsakis (changes requested):

System625:
My bad, thread-name was the wrong mechanism. Fixed: `record_coverage` is now `#[track_caller]` and uses `Location::caller()`, which chains through `ProvenSet::assert_ok` and `FormalityTest::ok` (both already `#[track_caller]`) so the recorded location points at the user's `#[test]` body. The JSONL identifier is now `test_file` + `test_line` + `test_column` instead of a thread-name string.

crates/formality-core/tests/coverage.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let file: PathBuf = tmp.join("test-coverage.jsonl");
+    let contents = std::fs::read_to_string(&file).expect("coverage file written");
+
+    assert!(
+        contents.contains("tests/coverage.rs"),
+        "missing test file in {contents}",
+    );
+    assert!(
+        contents.contains(&format!("\"test_line\":{all_even_line}")),
```
nikomatsakis, line 65: clever :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/361/73f406c98500e0382d567f72e037241b0bc5cc49..bc4a39f84dc52d2246d2d7795f0d9d5856262d95)*

nikomatsakis (approved):

PR #365, opened 2026-05-20
Improve duplication check for normal and negative impl

tiif:
## What does this PR do?

Previously, the duplication check for a normal impl ``impl A for B`` is not done in ``check_for_duplicate_items`` but in ``ensure_no_duplicate_impls`` under ``check_coherence``. It is a bit confusing and I think we can just do it in ``check_for_duplicate_items`` together with other crate items. 

For negative impl, we don't have any duplication check previously so I am just adding them in here. 

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/check/coherence.rs
+++ b/crates/formality-rust/src/check/coherence.rs
@@ -13,7 +13,6 @@ judgment_fn! {
 
         (
             (let current_crate_impls = program.trait_impls_in_crate(&current_crate))
-            (ensure_no_duplicate_impls(current_crate_impls) => ())
             (let all_crate_impls = program.trait_impls())
             (for_all(impl_a in current_crate_impls)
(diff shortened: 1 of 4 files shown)


lqd (approved):
:fire:

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/365/6a60712be5e5489b28a529a3a68312c88211b9d8..c5528dd948dea10cad55f5bd18a8359eb8abd95d)*

nikomatsakis (approved):
That makes sense to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/365/6a60712be5e5489b28a529a3a68312c88211b9d8..c5528dd948dea10cad55f5bd18a8359eb8abd95d)*

PR #366, opened 2026-05-21
Add formality-coverage crate to generate markdown coverage report

System625:
## What does this PR do?
Adds a new `formality-coverage` crate that generates a markdown coverage report from positive-test data recorded by #361.
Follow-up from the Zulip thread ["Coverage tracking: first PR plan"](https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/Coverage.20tracking.3A.20first.20PR.20plan).

- regex-scrapes the source tree for `judgment_fn!` blocks and `--- ("rule")` lines to enumerate every judgment/rule
- reads `target/test-coverage.jsonl` produced by positive tests
- emits `coverage.md` (top-level table) plus one subpage per judgment under `--out-dir`
- coverage marks are mdbook emoji shortcodes (`:check:` / `:x:`), matching the mockup on Zulip, so the mdbook preprocessor will render them
- ships a `formality-coverage` binary: `--src-root`, `--jsonl`, `--out-dir`

The new crate intentionally has no in-workspace dependencies (we don't reuse `RuleId` from `formality-core` as the JSONL reader only needs `(judgment, rule)` pairs and the scraper has its own types).

The scraper is intentionally hacky regex, as discussed on Zulip. It finds each `judgment_fn!` invocation, grabs the next `fn <name>(`, and treats everything up to the next `judgment_fn!` (or EOF) as that judgment's body so rule lines are matched with `^\s*-{3,}\s*\("([^"]+)"\)\s*$`. A rule-shaped line outside a macro invocation would be misattributed; the file-level comment notes this trade-off. The JSONL reader tolerates malformed lines (occasional interleaved parallel appends) with a warning rather than failing the report which is fine :)

## AI disclosure

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -358,6 +358,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "formality-coverage"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "expect-test",
+ "regex",
+ "serde_json",
+ "walkdir",
+]
+
+[[package]]
 name = "formality-macros"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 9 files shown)


nikomatsakis (approved):
Seems good to me! We should, as a follow-up PR, refactor formality-mdbook to depend on formaliy-coverage and share some of this logic. We might also want to embed these reports into the mdbook.

nikomatsakis (approved):

PR #369, opened 2026-05-27
Codegen, take 3

nikomatsakis:
## What does this PR do?

Next draft of codegen. Rebased.

It includes a chapter in the mdbook that explains the key ideas. The most important one is a `SemeRegion` (single-entry, multi-exit) that contains a fragment of a CFG. The `codegen_xxx` judgments convert bits of formality's Expr IR into `SemeRegion` output, these regions can then be combined. This is a nice setup in that you can convert expressions and things in a "standalone" fashion.

We do still thread a global state through everything to let us make globally unique identifiers; I don't love that, might be nice to switch away from it. 

## AI disclosure


BennoLossin:
do you mind splitting the codegen work from the toolchain fixes and merge those in first?

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -298,6 +298,19 @@ judgment_fn! {
...
+            (borrow_check_expr(
+                env,
+                assumptions,
+                state,
+                expr,
+                places_live_on_exit
+            ) => (_expr_ty, state))
+            ------------------------------------------------------------ ("print")
+            (borrow_check_statement(env, assumptions, state, Stmt::Print { expr }, places_live_on_exit) => (env, state))
+        )
```
tiif, line 314: what is the motivation to support print statement? (it's good to have of course, but i am just curious how is it related to codegen).

crates/formality-rust/src/check/impls.rs
tiif, line 1: is drop impl something related to minirust codegen specifically?

crates/minirust-rs/src/mem.rs
tiif, line 1: heh are we cloning the the minirust codebase here? other than running the script, is there a better way to do this?(maybe [josh](https://github.com/josh-project/josh) would help?)

crates/formality-rust/src/prove/prove/prove/prove_outlives.rs
```diff
@@ -52,6 +52,18 @@ judgment_fn! {
             (prove_outlives(_decls, _env, _assumptions, LtData::Static, _b) => Constraints::none(env))
         )
 
+        // Everything outlives 'erased
+        (
+            ----------------------------- ("anything outlives erased")
+            (prove_outlives(_decls, _env, _assumptions, _a, LtData::Erased) => Constraints::none(env))
+        )
```
tiif, line 59: mm why everything outlives ``'erased``?

tiif (commented):
very briefly read through the pr, haven't gotten to anything in ``codegen/*`` yet, we can dive  into that in the meeting

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/369/cc48e5d27171ad19bbb0b99653f43de7971cb026..8414a72ba9e42249784953caa120f827df13d6b7)*

crates/formality-rust/src/check/impls.rs
tiif: From what I understand from yesterday's meeting, the current drop impl support is incomplete, but niko has future plan for it. I think we can split this part to a separate PR if it is not used anywhere.

crates/formality-rust/src/codegen/code_block.rs
```diff
@@ -0,0 +1,633 @@
...
+    ///   y = tmp2;
+    /// ]
+    /// ```
+    /// 
+    /// and then you append those two so you get
+    /// 
+    /// ```
+    /// stmts = [... /* as above, concatenated */]
+    /// ```
+    Block { stmts: Vec<lang::Statement> },
```
tiif, line 46: would it be correct to say a ``Block`` (anonymous) will eventually get coverted to either ``Open`` or ``Closed``?

crates/formality-rust/src/codegen/code_block.rs
```diff
@@ -0,0 +1,633 @@
...
+                    fallthrough_stmts: Vec::new(),
+                },
+                cfn,
+            )
+        } else {
+            (CodeBlock::Closed { blocks, entry }, cfn)
+        }
+    }
+
+    fn region_entry_or_alloc(cfn: &CodegenFn, code: &CodeBlock) -> (lang::BbName, CodegenFn) {
```
tiif: This supposed to be ``block`` instead of ``region``?

crates/formality-rust/src/codegen/code_block.rs
```diff
@@ -0,0 +1,633 @@
...
+        match code {
+            CodeBlock::Block { .. } => cfn.fresh_bb(),
+            CodeBlock::Open { entry, .. } => (*entry, cfn.clone()),
+            CodeBlock::Closed { entry, .. } => (*entry, cfn.clone()),
+        }
+    }
+
+    fn drain_region_into(
```
tiif: This supposed to be ``block`` instead of ``region``?

crates/formality-rust/src/codegen/test.rs
tiif, line 1: we probably can remove this?

crates/formality-rust/src/codegen/scope.rs
```diff
@@ -0,0 +1,273 @@
...
+///
+/// The variable-to-local mapping (`vars`) is duplicated with `FlowState::scopes`
+/// in the sense that both know which variables are in scope and their types, but
+/// `vars` additionally maps each variable to its MiniRust local name, which
+/// `FlowState` doesn't track.
+///
+/// `FlowState` must be kept in sync as we codegen: when a new variable is introduced
+/// (`push_var`), it is registered in both `vars` and `FlowState`. When we call into
+/// the borrow-checker (e.g., `resolve_place`), we pass the current `flow_state` so
+/// it can verify liveness and loan validity at this program point.
```
tiif, line 76: So if I understand this correctly, we need ``FlowState`` because we do codegen and borrowck at the same stage / concurrently?

tiif (approved):
Thanks for working on this, it's very interesting to review :)

I briefly read through everything once, and I just have a few more questions. None of this is blocking, so feel free to merge with or without resolving the comments. 

Also, the PR description is still using ``SemeRegion``. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/369/cc48e5d27171ad19bbb0b99653f43de7971cb026..ff003c757d1b6fc87ca68c9c0c3b72f00eef7689)*

crates/formality-rust/src/check/borrow_check/nll.rs
nikomatsakis: minirust has a print -- it's the only way to get data *out* really

crates/minirust-rs/src/mem.rs
nikomatsakis: Yes, and that is intentional. It's "vendored", that way we can rebase periodically. If minirust starts getting published to crates.io we could use that.

crates/formality-rust/src/prove/prove/prove/prove_outlives.rs
nikomatsakis: beacuse we donj't know what `'erased` is, it's kind of "by definition we ignore stuff about it"

crates/formality-rust/src/codegen/code_block.rs
nikomatsakis: yes

crates/formality-rust/src/codegen/code_block.rs
nikomatsakis: yes

crates/formality-rust/src/codegen/code_block.rs
nikomatsakis: yes

crates/formality-rust/src/codegen/scope.rs
nikomatsakis: yes

PR #371, opened 2026-05-29
Lower `forall` and `outlives`

liolin:
## What does this PR do?

Lowers `forall` and `outlives` bounds. The implicit `Self` type in a trait definition is emitted correctly.
Modified tests to use `expect-test`.

Introduced a new `Wrapped` type that works as a RAII guard. By calling `Context::open_bounded` / `Context::open_bounded`, a fresh set of variable names is pushed to the context and the guard is returned. When `Wrapped` goes out of scope, the corresponding variable names are removed from the scope.

## AI disclosure

* Used AI for improving my doc comments.

### DIFF
--- a/crates/formality-rust/src/to_rust/context.rs
+++ b/crates/formality-rust/src/to_rust/context.rs
@@ -69,7 +69,7 @@ impl Context {
 
     /// Opens the binder `b` and instatiates ith with a fresh set of
     /// existential variables and returns the term.
-    pub fn open_exists<T: Term>(&mut self, b: Binder<T>) -> T {
+    pub fn open_exists<T: Term>(&mut self, b: Binder<T>) -> Wrapped<'_, T> {
         let subst = self.existential_substitution(&b);
         let term = b.instantiate_with(&subst).expect("suitable substitution");
         // TODO: How can I prevent the collect()?
@@ -82,8 +82,10 @@ impl Context {
                 )
             })
             .collect();
+        self.bounded
+            .insert(0, names.iter().cloned().map(|(_, name)| name).collect());
         self.free.extend(names.into_iter());
-        term
+        Wrapped::new(self, term.clone())
     }
 
     /// Allocates a new set of fresh names and returns a wrapped `T`.
--- a/crates/formality-rust/src/to_rust/context.rs
+++ b/crates/formality-rust/src/to_rust/context.rs
@@ -70,6 +70,7 @@ impl Context {
     /// Opens the binder `b` and instatiates ith with a fresh set of
     /// existential variables and returns the term.
     pub fn open_exists<T: Term>(&mut self, b: Binder<T>) -> Wrapped<'_, T> {
+        // TODO: Replace Existential Lifetimes with an Erased after codegen (#369) is merged
         let subst = self.existential_substitution(&b);
         let term = b.instantiate_with(&subst).expect("suitable substitution");
         // TODO: How can I prevent the collect()?
--- a/crates/formality-rust/src/to_rust/context.rs
+++ b/crates/formality-rust/src/to_rust/context.rs
@@ -13,6 +13,7 @@ pub struct Context {
     counter: VarIndex,
     bounded: Vec<Vec<String>>,
     free: HashMap<VarIndex, String>,
+    nesting_level: usize,
 }
 
 impl Context {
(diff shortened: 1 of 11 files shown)


crates/formality-rust/src/to_rust/context.rs
```diff
@@ -20,61 +25,110 @@ impl Default for Context {
...
+    /// these names are in scope. When the `Wrapped` value is dropped, the names
+    /// are removed from the contex
+    pub fn open_bounded<T: Term>(&mut self, b: Binder<T>) -> Wrapped<'_, T> {
+        let term = b.peek();
+        let names = b
+            .kinds()
             .into_iter()
-            .map(|var| format!("{}{}", self.kind_to_string(&var.kind), var.var_index.index))
+            .enumerate()
+            .map(|(index, kind)| format!("{}{}", self.kind_to_string(kind), index))
```
nikomatsakis: My only concern here -- if there is more than one level of binder (e.g., around the impl, and then, within that, around the method) -- won't this create shadowing?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/371/b786a52f5967523543cc967d93949e6cc4f5dce4..9ec083dc31d16325062bc74ef5be1d5dec27fad2)*

crates/formality-rust/src/to_rust/expr.rs
```diff
@@ -71,7 +74,10 @@ pub fn lower_let(
 
 pub fn lower_exists_stmt(ctx: &mut Context, binder: &Binder<Block>) -> Fallible<syntax::Stmt> {
     // TODO: Check again, after codegen is merged, if "erased" lifetimes could work here.
-    let (term, _) = ctx.open_exists(binder);
+    let Wrapped {
+        ref mut ctx,
+        ref term,
+    } = ctx.open_exists(binder.clone());
```
nikomatsakis, line 81: I think we should replace exists with "erased" lifetimes and render them as `'_` in the output, actually.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/371/b786a52f5967523543cc967d93949e6cc4f5dce4..9ec083dc31d16325062bc74ef5be1d5dec27fad2)*

nikomatsakis (changes requested):

crates/formality-rust/src/to_rust/expr.rs
liolin: The generated Rust code emits now `'_`. Because your code gen branch is not yet merged, the "erased" lifetimes are not available for me now. Will update, when code gen lands.

crates/formality-rust/src/to_rust/context.rs
liolin: You are right. I fixed this, tracking the nesting.
```rust
trait Foo<T> {                    
    fn blub<K, V>(k: K, v:V) -> T;
}
```
becomes
```rust
pub trait Foo<T01> {                                 
    fn blub<T10, T11>(mut k: T10, mut v: T11) -> T01;
}
```

lqd:
Same as you, this doesn't seem to reproduce locally...

Can you rebase now that the toolchain has been updated on main? We'll see if that changes things as well.

liolin:
Thank you for testing. I'll rebase

liolin:
With the rebase, I can reproduce it locally, because itertools are no longer a dependency. Does the CI somehow pick up a different `Cargo.toml` file?

lqd:
No idea. I imagine it shouldn't but there's still the question of what base is it using when testing: the PR's or whatever main is (just to make sure the PR would be green on merge?). That could have the same effect I guess, though I don't know when you started seeing the error.

lqd:
Make sure you add the dependency in a commit before starting using it in later commits, or we'd have commits that don't build and break bisection.

liolin:
@nikomatsakis I rebased the changes and added a commit ([84f39f0](https://github.com/rust-lang/a-mir-formality/pull/371/commits/84f39f0d9e582beadaa1d8cfb710a36be9419c2b)) to fix the itertools problem.

I don't use the `'erased` lifetime yet. I would prefer to add this in a new PR.

nikomatsakis (approved):

Issue #296, opened 2026-03-30
Track initialization and moves in the borrow checker


nikomatsakis:
Updates #254. Related to #209.

The borrow checker currently does not track whether variables are initialized or have been moved. This means programs that use uninitialized variables or use variables after they've been moved are accepted when they should be rejected.

For example, this should be an error but currently passes:

```
fn foo() -> u32 {
    let x: u32;
    return x;  // x was never initialized
}
```

And this should also be an error:

```
fn foo() -> u32 {
    let x: u32 = 0_u32;
    let y: u32 = x;   // x moved here
    let z: u32 = x;   // use of moved value
    return z;
}
```

**Ignored tests** (`#[ignore]` in `src/test/borrowck.rs`):

- `use_of_uninitialized_variable` — read from a never-initialized variable
- `use_of_moved_variable` — read from a variable after it was moved
- `conditional_init_one_branch` — variable initialized in only one if-branch, used after
- `assign_field_of_uninitialized` — assign to a field of an uninitialized struct (rustc rejects this)
- `partial_move_use_whole` — use whole struct after one field was moved
- `move_same_field_twice` — move the same field twice
- `move_whole_then_access_field` — access a field after the whole struct was moved
- `move_parent_then_access_child` — access `x.foo.bar` after `x.foo` was moved
- `move_in_loop` — move inside a loop body (visible on second iteration)
- `uninitialized_return` — return an uninitialized value (#209)

**Passing tests** (happy paths that should continue to pass):

- `reinit_after_move` — assignment after move re-initializes
- `conditional_init_both_branches` — both if-branches initialize
- `partial_move_use_sibling` — sibling field still accessible after partial move

PR #372, opened 2026-06-01
fix: track init and move in borrowck

phraakture:
## What does this PR do?

This pr is the implementation of approach-1 in the zulip chat, similar what we do in the compiler. This adds a `Set<PlaceExpr>` to the flow state that tracks which places are uninitialized. It's a forward analysis in terms of approach-2(liveness) and it uses union at joins.

related issue #296

Tracks `Set<PlaceExpr>` of uninitialized places in `PointFlowState.uninit`. Union at control-flow joins. 
On access: `check_place_initialized` fails if any prefix or sub-path is in the set. 
On assignment: `check_place_writable` fails if a strict prefix is in the set, then marks LHS initialized. 
On move: marks the place `uninit`. 
On no init: marks `uninit`. 

Questions:
1. `move_in_loop` still ignored- some direction on handling that would be appreciated

## AI disclosure
* used to understand the codebase, improve naming and to keep comments consistent to the codebase


phraakture:
r? nikomatsakis

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -66,6 +66,10 @@ pub struct PointFlowState {
 
     /// The set of loans that are live (issued, not killed)
     pub loans_live: Set<Loan>,
+
+    /// Places that are uninitialized or have been moved from
+    /// Moving from any place whose prefix is in this set is an error
```
tiif:
```suggestion
    /// Read or write or move from any place whose prefix is in this set is an error
```

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -79,6 +83,23 @@ impl PointFlowState {
...
             outlives: Union((&self.outlives, outlives)).upcast(),
             loans_live: self.loans_live.clone(),
+            uninit: self.uninit.clone(),
+        }
+    }
+
+    /// Mark a place as initialized: remove it and all sub-paths from uninit
+    pub fn mark_initialized(&mut self, place: &PlaceExpr) {
+        self.uninit.retain(|u| !place.is_prefix_of(u));
+    }
```
tiif, line 98: This is more of a personal preference, would it be possible to use ``filter`` instead of ``retain``?

I think it might be clearer to say ``when a place is initialized, remove it and all other sub path from uninit`` through ``filter`` instead of ``retain anything that is *not* initialized``.

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -203,6 +227,18 @@ impl FlowState {
...
+        this.current.mark_initialized(place);
+        this
+    }
+
+    pub fn with_uninit(&self, place: &PlaceExpr) -> Self {
+        let mut this = self.clone();
+        this.current.mark_uninit(place);
+        this
+    }
+
```
tiif, line 251: would it be better if we just call ``this.current.mark_initialized``directly at the call site?

crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -423,6 +426,8 @@ judgment_fn! {
                 place,
             ) => (place, state))
 
+            // Check that the accessed place is initialized
+            (if check_place_initialized(&state, &place.to_place_expression()))
```
tiif: Maybe it will be better to move ``check_place_initialized`` and ``check_place_writable``  into ``access_permitted`` 🤔

tiif (commented):
Thank you for your contribution! I left some suggestions. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/372/b786a52f5967523543cc967d93949e6cc4f5dce4..52f67a56ec8eaf36bb1a6a35de9124819beaf2cd)*

crates/formality-rust/src/check/borrow_check/flow_state.rs
phraakture: yeah we can do that, `set` is `BTreeSet`, as ther's no filter method on `BTreeSet`, we just need to rebuild on iterator

crates/formality-rust/src/check/borrow_check/flow_state.rs
phraakture: maybe, idk, i think as they follow the same pattern as `with_loan`, `with_outlives`, `with_break`, etc its bit consistent to the codebase. Also its not costing us anything either way.

if you still prefer it, it's pretty straightforward ig and i can take the stab

nikomatsakis (commented):
I'll give this one more read but I feel quite good about it. Nice work @phraakture, I'm sure we'll continue to evolve, but it seems good to be closing this gap.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/372/b786a52f5967523543cc967d93949e6cc4f5dce4..4ba31ea3737225622a7e12d98723af6661cd4749)*

nikomatsakis (approved):
I'd like to merge this -- I see it has merge conflicts though :(

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/372/b786a52f5967523543cc967d93949e6cc4f5dce4..4ba31ea3737225622a7e12d98723af6661cd4749)*

tiif:
Hi @phraakture, could you rebase this so we can merge it?

phraakture:
@nikomatsakis @tiif rebased, ready for merge i think

PR #375, opened 2026-06-03
Remove unused deps

bit-aloo:
## What does this PR do?

We have a couple unused dependencies, which this PR removes. The major motivation was this thread: https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/target.20directory.20size/with/599523951, but unfortunately this removal is not that detrimental like it only reduced target size by 100 MB

Before:

```sh
du -sh target/
3.1G    target/
```
After:

```sh
du -sh target/
3.0G    target/
```

An effort to reduce the target directory size, which brought it down by around 100 MB. I used `cargo machete` for this. I thought about adding it to CI, but for some reason we don’t currently have CI checks even for clippy/fmt, so I didn’t add it yet. Though I would like to.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,7 +12,6 @@ dependencies = [
  "formality-core",
  "formality-macros",
  "formality-rust",
- "pretty_assertions",
  "tracing",
 ]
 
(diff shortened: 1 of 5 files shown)


lqd (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/375/cc48e5d27171ad19bbb0b99653f43de7971cb026..d42c1ddde5085f8317953d5e896eefaaa50eb768)*

bit-aloo:
@lqd, should we add machete to CI along with clippy and fmt? Also, is there a reason why we don’t already have them?

lqd:
I don't think we need machete on CI per se, having unused dependencies is not that common of an occurrence. We already have rustfmt though.

bit-aloo:
> I don't think we need machete on CI per se, having unused dependencies is not that common of an occurrence. We already have rustfmt though.

Clippy?

lqd:
We have a lot of macros and that usually tends to trigger false positives there. We may not need it yet, while things are moving a lot.

PR #394, opened 2026-06-08
Coverage implicit no match

System625:
## What does this PR do?

Follow-up to #378. Refines negative coverage on two axes Niko flagged on
2026-06-08: Implicit no-match tracking and Infallible-premise tagging. The `(exhaustive)` rule marker is deferred to a separate PR since it needs macro and runtime changes. Stacked on #378.

When a judgment fails with no rule blamed (either no conclusion pattern matched the input, or every match got stripped before its `!` clause), we now record the judgment itself. The report surfaces it on the judgment row and at the top of the subpage. 
The scraper classifies each premise (`let`, `if`, `if let`, judgment call) and applies Niko's heuristic: `let` is infallible unless its body contains `?`. Rules whose premises are all infallible cannot be blamed once their conclusion matches, so the report renders their negative cell as `N/A` rather than flagging it as a coverage gap.

## AI disclosure


crates/formality-core/src/judgment/coverage.rs
```diff
@@ -99,11 +103,13 @@ pub fn record_negative_coverage<'a>(failures: impl IntoIterator<Item = &'a Faile
     let caller = Location::caller();
 
     let mut rules: BTreeSet<BlamedRule> = BTreeSet::new();
+    let mut implicit: BTreeSet<ImplicitNoMatch> = BTreeSet::new();
     for failure in failures {
         rules.extend(failure.collect_blamed_rules());
+        implicit.extend(failure.collect_implicit_no_match());
```
nikomatsakis: Hmm. I'm not convinced that these should be separate variables. I expected that we would just have one set of "negatively tested rules".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/394/ed29de4773cffa361a18e37397ca015da5b31ef5..331a30a63283b740579bf1f5b9b89b8859b3a0e6)*

crates/formality-core/src/judgment/coverage.rs
```diff
@@ -201,6 +211,20 @@ fn format_negative_jsonl(caller: &Location<'_>, rules: &BTreeSet<BlamedRule>) ->
         push_json_str(&mut s, r.cause);
         s.push('}');
     }
+    s.push_str("],\"implicit_no_match\":[");
```
nikomatsakis: pre-existing but we should rewrite this to use `serde_json`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/394/ed29de4773cffa361a18e37397ca015da5b31ef5..331a30a63283b740579bf1f5b9b89b8859b3a0e6)*

crates/formality-core/src/judgment/proven_set.rs
```diff
@@ -617,6 +617,49 @@ impl FailedJudgment {
...
+    /// Recorded by negative coverage so the report can mark a judgment as
+    /// "exercised but unmatched" even when there's no specific rule to
+    /// blame.
+    pub fn collect_implicit_no_match(&self) -> BTreeSet<ImplicitNoMatch> {
+        let mut acc = BTreeSet::new();
+        self.collect_implicit_no_match_into(&mut acc);
+        acc
+    }
+
+    fn collect_implicit_no_match_into(&self, acc: &mut BTreeSet<ImplicitNoMatch>) {
```
nikomatsakis: As above, I think we can just combine this with the notion of "blame" and just record every rule along the stack trace. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/394/ed29de4773cffa361a18e37397ca015da5b31ef5..331a30a63283b740579bf1f5b9b89b8859b3a0e6)*

nikomatsakis: It seems to me that what we want is to say something like -- we have a failure that failed "because of" --

* judgment J failing to prove premise P of rule R
* judgment J failing to find an applicable rule (always the leaf)

And we just want to collect a set of those.

crates/formality-coverage/src/jsonl.rs
```diff
@@ -90,7 +109,12 @@ pub fn read(path: &Path) -> Result<Coverage> {
             .and_then(|v| v.as_str())
             .unwrap_or("positive");
         match kind {
```
nikomatsakis: Again, rewrite this to use serde and be well-typed. It should make use of shared struct definitions between the serializer + deserializer.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/394/ed29de4773cffa361a18e37397ca015da5b31ef5..331a30a63283b740579bf1f5b9b89b8859b3a0e6)*

crates/formality-coverage/src/scrape.rs
```diff
@@ -29,6 +29,33 @@ pub struct Rule {
...
+/// uncovered cell renders as N/A rather than a gap.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Premise {
+    pub raw_text: String,
+    pub kind: PremiseKind,
+    pub fallible: bool,
+}
+
+/// Coarse classification of premise syntax. Drives the fallibility heuristic
+/// Niko outlined: `let` is infallible unless its body contains `?`; `if`,
```
nikomatsakis: This is an "AI" comment -- it is documenting our conversation, not the code

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/394/ed29de4773cffa361a18e37397ca015da5b31ef5..331a30a63283b740579bf1f5b9b89b8859b3a0e6)*

nikomatsakis (changes requested):

nikomatsakis (approved):

PR #398, opened 2026-06-14
Add feature gate for non lifetime bound

### LINKED ISSUE #379: Discrepancy in test case `basic_where_clauses_pass`
The test case [basic_tests::basic_where_clauses_pass](https://github.com/rust-lang/a-mir-formality/blob/0ab4832e50e95cf4aa706a26f189299829a37e60/tests/basic_tests.rs#L66-L77) passes type and borrow check in formality but fails in rustc, because rustc only allows lifetime parameters in an for all statement:

```rust
thread 'basic_where_clauses_pass' (69934) panicked at tests/basic_tests.rs:80:6:
expected `rustc` to succeed but it failed:
    Checking mycore v0.1.0 (/tmp/formality-AfDPLq/crates/mycore)
error[E0658]: only lifetime parameters can be used in this context
  --> crates/mycore/src/lib.rs:11:9
   |
11 |     for<T0> u32: A<T0>,
   |         ^^
   |
   = note: see issue #108185 <https://github.com/rust-lang/rust/issues/108185> for more information
   = help: add `#![feature(non_lifetime_binders)]` to the crate attributes to enable
   = note: this compiler was built on 2025-11-30; consider upgrading it if it is out of date

For more information about this error, try `rustc --explain E0658`.
error: could not compile `mycore` (lib) due to 1 previous error
```

josetorrs:
## What does this PR do?

Fixes https://github.com/rust-lang/a-mir-formality/issues/379. Closed https://github.com/rust-lang/a-mir-formality/pull/397 because I realized that's off my local main 🤦‍♂️ 

Think title and the issue is self-explanatory but this checks if the `non_lifetime_bound` feature gate is there whenever lifetime parameters are used and rejects if it's not

I do have a question if the below matters at all? I looked to the surrounding `check_for_duplicate_items` for alot of inspiration and guidance on this. 
```rust
fn check_for_non_lifetime_binders(p: &Program) -> Fallible<ProofTree> {
    let Crates { crates } = p.program();
    //.. 
```

## AI disclosure

* Other: (explain)

I had Claude initially point me to the relevant files, then looked at the surrounding code and worked on it. Finally, at the end I asked if it had any suggestions after looking at the issue but for the most part it just agreed with the choices but that might just be an LLM being an LLM.

I did ask how to run it using rustc and it said to use `FORMALITY_RUN_RUSTC` (which worked) then pointed to the comment [here](https://github.com/rust-lang/a-mir-formality/blob/1f66a4773cb3c365efec27d801b9c66d71a48ab4/src/test_util.rs#L63)  but I do think it might be helpful to add it to the book [here](https://rust-lang.github.io/a-mir-formality/intro.html?search=#running-tests) because at first I was a little confused on how to reproduce it

### DIFF
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -149,7 +149,6 @@ fn check_for_non_lifetime_binders(c: &Crate) -> Fallible<ProofTree> {
                     bail!("non lifetime binders require #![feature(non_lifetime_binders)]");
                 }
             }
-            // TODO: impl, fn, adt, ..
             _ => {}
         }
     }
--- a/crates/formality-rust/src/check/mod.rs
+++ b/crates/formality-rust/src/check/mod.rs
@@ -6,8 +6,7 @@ use crate::prove::prove::{is_definitely_not_proveable, Constraints, Env, Program
 use crate::rust::Visit;
 use crate::{
     grammar::{
-        Crate, CrateId, CrateItem, Crates, Fallible, FeatureGateName, Test, TestBoundData,
-        Wcs,
+        Crate, CrateId, CrateItem, Crates, Fallible, FeatureGateName, Test, TestBoundData, Wcs,
     },
     prove::ToWcs,
 };
@@ -136,7 +135,7 @@ fn check_for_non_lifetime_binders(c: &Crate) -> Fallible<ProofTree> {
     );
 
     if is_non_lifetime_binder_enabled {
-        return Ok(ProofTree::leaf("check_for_non_lifetime_binders: ok"));
+        return Ok(ProofTree::leaf("check_for_non_lifetime_binders: feature enabled"));
     }
 
     if c.items.iter().any(item_has_non_lifetime_bind) {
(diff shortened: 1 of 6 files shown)


crates/formality-rust/src/check/mod.rs
```diff
@@ -120,6 +123,33 @@ fn check_for_duplicate_items(program: &Program) -> Fallible<ProofTree> {
...
 
+fn check_for_non_lifetime_binders(c: &Crate) -> Fallible<ProofTree> {
+    let is_non_lifetime_binder_enabled = c.items.iter().any(|item|
+        matches!(item, CrateItem::FeatureGate(fg) if fg.name == FeatureGateName::NonLifetimeBinders)
+    );
+
+    for item in c.items.iter() {
+        match item {
+            CrateItem::Trait(t) => {
+                let has_non_lifetime_binder = t
```
nikomatsakis: This is good! But also, there are other kinds of items (functions, structs, etc) where should be doing similar checks. Can you extend with those (and tests thereof?)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/398/1f66a4773cb3c365efec27d801b9c66d71a48ab4..7b926ec3d41ce382d09b68b2ce3b2561ad069716)*

nikomatsakis (changes requested):

crates/formality-rust/src/check/mod.rs
josetorrs: added mostly in commit: [83d0390](https://github.com/rust-lang/a-mir-formality/pull/398/commits/83d0390a8eb37ea172ec15a71d1bc000f8cf1af9)

nikomatsakis (approved):

PR #402, opened 2026-06-18
Make sure to consider trait ref projection unknowanle

### LINKED ISSUE #387: Discrepancy in test case `is_local_unknowable_trait_ref`
The test case [coherence_overlap::is_local_unknowable_trait_ref](https://github.com/rust-lang/a-mir-formality/blob/0ab4832e50e95cf4aa706a26f189299829a37e60/tests/coherence_overlap.rs#L282-L304) passes in formality but fails in rustc.

I get the following error when compiling with rustc:

```
thread 'is_local_unknowable_trait_ref' (96968) panicked at tests/coherence_overlap.rs:317:6:
expected `rustc` to succeed but it failed:
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Checking mycore v0.1.0 (/tmp/formality-Tv4YpF/crates/mycore)
    Checking foo v0.1.0 (/tmp/formality-Tv4YpF/crates/foo)
error[E0119]: conflicting implementations of trait `Overlap<LocalType>` for type `()`
 --> crates/foo/src/lib.rs:9:1
  |
7 | impl<T0, T1> Overlap<T1> for T0 where <T0 as Project>::Assoc: Foo<T1> {}
  | --------------------------------------------------------------------- first implementation here
8 |
9 | impl<T0> Overlap<LocalType> for () {}
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `()`

error[E0207]: the type parameter `T0` is not constrained by the impl trait, self type, or predicates
 --> crates/foo/src/lib.rs:9:6
  |
9 | impl<T0> Overlap<LocalType> for () {}
  |      ^^ unconstrained type parameter

Some errors have detailed explanations: E0119, E0207.
For more information about an error, try `rustc --explain E0119`.
error: could not compile `foo` (lib) due to 2 previous errors
```

bit-aloo:
## What does this PR do?
closes: #387 

We have this discrepancy with the rustc in case if we have a remote trait ref with a projection, on formality side we would consider such a case to be ok considering some downstream might implement it, whereas in rustc we consider it unknowable. So, here we are now doing the same for formality, like checking if it is foreign trait reference and if it is we then see it is an alias and then we don't allow it. I am not entirely confident with this but it does not break any test fortunately :).

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/prove/prove/prove/is_local.rs
+++ b/crates/formality-rust/src/prove/prove/prove/is_local.rs
@@ -1,6 +1,7 @@
 use crate::{
     grammar::{
-        AliasTy, Lt, Parameter, PredicateTy, RigidName, RigidTy, TraitRef, TyData, Variable, Wcs,
+        AliasName, AliasTy, Lt, Parameter, PredicateTy, RigidName, RigidTy, TraitRef, TyData,
+        Variable, Wcs,
     },
     prove::prove::Constrained,
 };
@@ -58,7 +59,6 @@ judgment_fn! {
     ) => Constraints {
         debug(assumptions, goal, env)
         assert(env.bias() == Bias::Completeness)
-
         (
             (may_be_downstream_trait_ref(decls, env, assumptions, goal) => c)
             --- ("may be defined downstream")
@@ -104,6 +104,13 @@ judgment_fn! {
     ) => Constraints {
         debug(parameter, assumptions, env)
         assert(env.bias() == Bias::Completeness)
+        (
+            // Projections from remote traits are not safe to rule out during
+            // coherence, so we treat them as maybe downstream.
+            (if !decls.is_local_trait_id(&name.as_associated_ty_id().unwrap().trait_id))
+            --- ("projection from remote trait")
+            (may_be_downstream_parameter(decls, env, assumptions, AliasTy { name, parameters: _ }) => Constraints::none(env))
+        )
         (
             // existential variables *could* be inferred to downstream types; depends on the substitution
             // we ultimately have.
--- a/tests/coherence_overlap.rs
+++ b/tests/coherence_overlap.rs
@@ -304,22 +304,11 @@ fn is_local_unknowable_trait_ref() {
         impl Overlap<LocalType> for () {}
     }])
     .skip_execute()
-    .rustc_err(expect_test::expect![[r#"
-            Checking mycore v0.1.0 (/tmp/formality-NXqOxy/crates/mycore)
-            Checking foo v0.1.0 (/tmp/formality-NXqOxy/crates/foo)
-        error[E0119]: conflicting implementations of trait `Overlap<LocalType>` for type `()`
-         --> crates/foo/src/lib.rs:9:1
-          |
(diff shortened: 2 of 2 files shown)


crates/formality-rust/src/prove/prove/prove/is_local.rs
```diff
@@ -58,7 +59,15 @@ judgment_fn! {
...
         assert(env.bias() == Bias::Completeness)
-
+        (
+            // If a remote trait ref contains a projection/alias, we treat it
+            // conservatively during coherence. Like instead of proving that no
+            // unknown crate could have it, we treat the trait ref as unknowable.
+            (if !decls.is_local_trait_id(&goal.trait_id))
+            (if trait_ref_has_alias_parameter(&goal))
+            --- ("may be defined upstream via alias")
+            (may_be_remote(decls, env, assumptions, goal) => Constraints::none(env))
```
lcnr: this should be a part of `may_be_downstream_trait_ref` 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/402/ec50f56846d52768cfae15be19f2dc67d71455b8..166fe182ef9731d06aa7d8a477b31d1ec0550baf)*

bit-aloo: we do have `may_be_downstream_parameter`, I added the check there. Thanks.

crates/formality-rust/src/prove/prove/prove/is_local.rs
```diff
@@ -104,6 +104,13 @@ judgment_fn! {
     ) => Constraints {
         debug(parameter, assumptions, env)
         assert(env.bias() == Bias::Completeness)
+        (
+            // Projections from remote traits are not safe to rule out during
+            // coherence, so we treat them as maybe downstream.
+            (if !decls.is_local_trait_id(&name.as_associated_ty_id().unwrap().trait_id))
+            --- ("projection from remote trait")
+            (may_be_downstream_parameter(decls, env, assumptions, AliasTy { name, parameters: _ }) => Constraints::none(env))
```
lcnr, line 112: can you explain how remote vs local traits differ here?

this is a rule that doesn't exist in rustc to my knowledge

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/402/ec50f56846d52768cfae15be19f2dc67d71455b8..95873f69d8f0fed2cde644c7fe0b0a5ad00c9665)*

bit-aloo: If a remote trait reference contains a projection or type alias, we flag it during coherence because the current crate does not own the upstream trait. As a result, it cannot fully reason about all the possible types that the projection or alias might resolve to.

For this reason, I only added checks for projections and aliases that come from upstream traits.

I think I may still be missing something here, and my reasoning is probably not explained very clearly.

bit-aloo:
I am closing this now, I think I need to think deeper about this, instead of this hack.

PR #403, opened 2026-06-23
Run tests with rustc

liolin:
## What does this PR do?

Runs part of the current a-mir-formality test suite with rustc enabled. I only updated the test modules, where the whole module passes with rustc. I'll add the remaining test modules, after discussion how to handle the discrepancy.

Add `rustc_ok()` or `rustc_err(...)` to the test cases.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-core/src/test_util.rs
+++ b/crates/formality-core/src/test_util.rs
@@ -46,8 +46,14 @@ pub fn normalize_paths(s: impl Display) -> String {
     let s = s.to_string();
     // Capture the filename (without path) and strip line:column
     // Handle both forward slashes (Unix) and backslashes (Windows) as path separators
+
+    // Matches paths in the formality output
     let re = regex::Regex::new(r"\((?:[^()]+[/\\])?([^()/\\]+\.rs):\d+:\d+\)").unwrap();
-    re.replace_all(&s, "($1)").to_string()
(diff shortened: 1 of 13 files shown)


lqd:
I wonder if we’ll need more than a single rustc result per test sometimes, e.g. the result with NLLs, the result with polonius, etc. We could split these into separate tests of course, but it’s interesting to think about if we don’t, and avoid duplication.

liolin:
The CI fails because Windows uses backslashes (`\`) as folder separator, which does not match with the `expect!` macro call. I have to find a solution that either normalizes the paths or omits the paths completely.

lqd:
It could be enough/easy to normalize the lines starting with `-->`?

liolin:
Good idea. I'll try this right now.

liolin:
I had finally the time to fix this. @nikomatsakis, can you please review it?

nikomatsakis (approved):
This looks great @liolin -- unfortunately it needs a rebase :(

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/403/c490389679ab0d790cfad83f20dccae46537589e..1df5bca54285d291c71904c500c498e27de18604)*

PR #404, opened 2026-06-23
Add per-cell coverage detail pages listing tests

System625:
## What does this PR do?

Follow-up to #401. Coverage cells now link to a per-cell **detail page** listing every test involved, instead of the ✓ linking only to the first test.

Each cell links to `./<slug>.md`, and the same `slug` helpers produce the detail pages, so links and pages always line up. The negative test locations were already in the JSONL; the reader just wasn't keeping them so changes there.

## The Coverage Table
image

## The Detail Page
image

## AI disclosure

### DIFF
--- a/crates/formality-coverage/src/report.rs
+++ b/crates/formality-coverage/src/report.rs
@@ -9,7 +9,7 @@
 //! with three columns (line number, coverage count, source line). The number on
 //! a rule's conclusion is positive coverage; the number on a premise is negative
 //! coverage. Both link to a per-cell detail page ([`render_detail_pages_for`])
-//! that lists the individual tests behind nested disclosure triangles.
+//! that lists each individual test with its source location and source.
 
 use crate::jsonl::{Coverage, TestLoc};
 use crate::scrape::{Judgment, Premise, Rule};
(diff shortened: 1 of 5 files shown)


System625:
3-column code view (line # | coverage count | source) instead of the old markdown tables. The number on a rule's conclusion is positive coverage; the number on each premise is negative coverage. Clicking a number opens that cell's detail page.
Disclosure triangles on the detail pages: an outer triangle reveals the tests, and each test is its own triangle revealing its source. The proof tree rendering will come in a follow-up PR

## New Subpage
image

## New Detail Page
image

nikomatsakis:
@System625 it'd be good to see 

(a) for the judgment, what is the "Signature"

and 

(b) for the test caess, I want to see the actual *content* of the text. I'm not sure if the disclosure triangles are useful if we are going to a separate page, maybe just show like

````
---
Source locaiton: foo/bar.rs:64

```
...test contents here...
```
````

File: crates/formality-coverage/src/report.rs [hunk 1]
```diff
@@ -305,40 +301,93 @@ pub fn render_detail_pages_for(
...
+    let indent = lines[fn_idx].len() - lines[fn_idx].trim_start().len();
+
+    // Include any attribute lines (e.g. `#[test]`) directly above the `fn`.
+    let mut start = fn_idx;
+    while start > 0 && lines[start - 1].trim_start().starts_with("#[") {
+        start -= 1;
+    }
+
+    let close = format!("{}}}", " ".repeat(indent));
+    let end = (fn_idx + 1..lines.len()).find(|&i| lines[i] == close)?;
```
nikomatsakis, line 376: Hmm. This is hacky but I suppose it works. A "nicer" alternative would be to find matching braces but... it's fine.  

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/404/1f19c5292deb80a7f469149a6591d89b5a3be49d..78951fc88a81db39d7dd1e2354d1165fc63edad7)*

nikomatsakis (approved):

PR #406, opened 2026-06-25
add: optional-fn-return-type

### LINKED ISSUE #393: Make function without return type nicer
Currently, the test below...
```rust
#[test]
fn i_am_a_test_that_should_be_rename_later() {
        fn foo() {
        }
    }])
    .ok()
}
```
can only be expressed with 
```rust
#[test]
fn i_am_a_test_that_should_be_rename_later() {
        fn foo() -> () {
        }
    }])
    .ok()
}
```

It will be nice to write the first test instead of the second one.

xonx4l:
## What does this PR do?
This PR implements optional function return type as discussed in https://github.com/rust-lang/a-mir-formality/issues/393

e.g., closes #393

Functions can now be written `fn foo() { }` instead of requiring`fn foo() -> () { }`. When the return type is omitted, it defaults to the unit type `()`.      

Added tests demonstrating that `fn run() {trusted}` parses identically to`fn run() -> () {trusted}`.
 
## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-macros/src/parse.rs
+++ b/crates/formality-macros/src/parse.rs
@@ -500,132 +500,60 @@ fn wrap_field_mode(
         }
 
         FieldMode::Guarded { guard, mode } => {
-            let guard_keyword = as_literal(guard);
-            match mode.as_ref() {
+            // A guard is present when its leading token matches. `cond` is a
+            // `bool` expression that consumes the leading token on success (and
+            // nothing on failure) `after` consumes any remaining guard tokens
+            // once we have committed to the guard.
+            let (cond, after) = match guard {
+                spec::Guard::Keyword(ident) => {
+                    let keyword = as_literal(ident);
+                    (quote!(__p.expect_keyword(#keyword).is_ok()), quote!())
+                }
+                spec::Guard::Operator(operator) => {
+                    let mut chars = operator.chars();
+                    let first = chars
+                        .next()
+                        .expect("operator guard must have at least one character");
+                    let rest: Vec<TokenStream> =
+                        chars.map(|c| quote!(__p.expect_char(#c)?;)).collect();
+                    (quote!(__p.expect_char(#first).is_ok()), quote!(#(#rest)*))
+                }
+            };
+
+            // How to parse the field once the guard has matched.
+            let parse_present = match mode.as_ref() {
                 FieldMode::Single => {
                     if let Some(ty) = field_ty {
-                        quote_spanned!(name.span() =>
-                            match __p.expect_keyword(#guard_keyword) {
-                                Ok(()) => {
-                                    __p.each_nonterminal(|#name: #ty, __p| {
-                                        #inner
-                                    })
-                                }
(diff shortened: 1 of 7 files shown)


crates/formality-macros/src/parse.rs
```diff
@@ -500,132 +500,60 @@ fn wrap_field_mode(
...
+                    (quote!(__p.expect_keyword(#keyword).is_ok()), quote!())
+                }
+                spec::Guard::Operator(operator) => {
+                    let mut chars = operator.chars();
+                    let first = chars
+                        .next()
+                        .expect("operator guard must have at least one character");
+                    let rest: Vec<TokenStream> =
+                        chars.map(|c| quote!(__p.expect_char(#c)?;)).collect();
+                    (quote!(__p.expect_char(#first).is_ok()), quote!(#(#rest)*))
```
nikomatsakis, line 519:
```suggestion
                    (
                        quote!(__p.expect_char(#first).is_ok()), 
                        quote!(#(__p.expect_char(#chars)?;)*)
                    )
```

I think maybe maybe you can just do this?

crates/formality-rust/src/grammar/ty.rs
```diff
@@ -94,6 +94,13 @@ impl Ty {
     }
 }
 
+/// The default type is the unit type `()`.
```
nikomatsakis, line 97: Hmm, I am not sure if we want a default for `Ty` that is always unit. It's convenient here but it might not be correct elsewhere. Let's have a new type like

```rust
struct OutputTy { ty: Ty }

impl Upcast<Ty> for OutputTy { }

impl Default for OutputTy { }
```

and then change `output_ty: OutputTy`

nikomatsakis (changes requested):
Nice work! I have one suggested improvement

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/406/1f19c5292deb80a7f469149a6591d89b5a3be49d..84015de123190fe4b04690040edf58f9765239c5)*

crates/formality-rust/src/grammar/ty.rs
xonx4l: Ok , will do . Thanks!

nikomatsakis (changes requested):
(Still waiting for change to have an `OutputTy` that carries the `Default` impl.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/406/1f19c5292deb80a7f469149a6591d89b5a3be49d..84015de123190fe4b04690040edf58f9765239c5)*

xonx4l:
@nikomatsakis I am sorry , I really got a bit busy and forgot about this . I will try to implement the suggested changes asap. Thank you!

nikomatsakis:
@xonx4l no worries!

xonx4l:
@nikomatsakis sorry it took a while , made all the requested changes . Thank you!

PR #408, opened 2026-06-29
Render per-test proof trees on coverage detail pages

System625:
## What does this PR do?

Renders per-test proof trees on the coverage detail pages (follow-up to #404). Positive cells show each test's success proof tree behind a collapsed disclosure; negative cells show the failed proof tree, pruned to the stacks that blame that premise plus their ancestors.

So the trees are stored in the JSONL as mirror structs not the real ProofTree. For positive trees I drop the if/let/trivial leaf nodes, so what's left is the rules that fired. For negative pages I prune each test's failed tree to only the stacks that reach the premise the page is about.

### Positive Proof Tree
image

### Negative Proof Tree
image
image

## AI disclosure

### DIFF
--- a/crates/formality-core/src/judgment/coverage.rs
+++ b/crates/formality-core/src/judgment/coverage.rs
@@ -20,7 +20,8 @@ use std::io::Write;
 use std::panic::Location;
 use std::path::PathBuf;
 
-use super::{FailedJudgment, FailureReason, ProofTree};
+use super::proven_set::judgment_name_prefix;
+use super::{FailedJudgment, FailedRule, FailureReason, ProofTree, RuleFailureCause};
 
 /// A single (judgment, rule) pair as identified in positive coverage output.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


crates/formality-core/src/judgment/coverage.rs
```diff
@@ -36,27 +37,139 @@ pub struct RuleId {
...
+        /// [`FailedTreeNode`]). Defaulted for forward compatibility, as above.
+        #[serde(default, skip_serializing_if = "Vec::is_empty")]
+        trees: Vec<FailedTreeNode>,
     },
 }
 
+/// A structural snapshot of a [`ProofTree`] node for embedding in positive
+/// coverage records. We keep the proof *shape* (judgment name, the rule that
+/// fired, the source location, and children) and drop the per-node attributes
+/// (argument/result `Debug` dumps), which are large and vary run-to-run.
```
nikomatsakis: Hmm. Certainly they can be large, but also, they are pretty interesting when you're browsing. We may want to revisit this. I don't think they actually vary from run-to-run in any notable way.

crates/formality-core/src/judgment/coverage.rs
```diff
@@ -36,27 +37,139 @@ pub struct RuleId {
...
+/// faithful tree of the rules that actually fired.
+///
+/// `rule` is owned because the live `ProofTree.rule_name` is `&'static str`,
+/// which serde can serialize but cannot deserialize into; a mirror type lets
+/// the report read trees back without changing the proving engine's hot type.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ProofTreeNode {
+    pub judgment: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rule: Option<String>,
```
nikomatsakis, line 140: What is this `None`?

crates/formality-core/src/judgment/coverage.rs
```diff
@@ -36,27 +37,139 @@ pub struct RuleId {
...
+/// Every node we keep is therefore a rule application or a `for_all`, all of
+/// which have small, stable names. This keeps the snapshot both lean and a
+/// faithful tree of the rules that actually fired.
+///
+/// `rule` is owned because the live `ProofTree.rule_name` is `&'static str`,
+/// which serde can serialize but cannot deserialize into; a mirror type lets
+/// the report read trees back without changing the proving engine's hot type.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ProofTreeNode {
+    pub judgment: String,
```
nikomatsakis, line 134: What does this string name? It'd be good to have an example

crates/formality-core/src/judgment/coverage.rs
nikomatsakis, line 14: I think it would be very helpful to have (at the module level) a standalone example of some judgment rules and two tests, one positive and one negative. That way in the rustdocs on the trees we can explain the values of the various variables and the resulting trees based on the example.

crates/formality-core/src/judgment/coverage.rs
```diff
@@ -182,11 +301,14 @@ fn coverage_dir() -> Option<PathBuf> {
...
-// have similar guarantees on Windows for ordinary files, so parallel `cargo test`
-// processes can safely append to the same file without locking, provided each
-// line stays small.
+// Parallel `cargo test` processes append to this file concurrently without
+// locking. Appends of a single payload are atomic for ordinary files, so we
+// build the record plus its trailing newline up front and emit it with one
+// `write_all` (rather than `writeln!`, which can issue separate writes for the
+// line and the newline and interleave under concurrency). The reader still
+// tolerates the occasional torn line, since embedding proof trees makes
+// individual lines larger and no atomicity guarantee is unconditional.
```
nikomatsakis: This sounds kinda like AI nonsense. What does it even mean ("no guarantee is unconditional"). It's either atomic or it's not. If it's not atomic, then we should use OS-level locking primitives.

nikomatsakis (changes requested):
Left some suggestions, mostly asking for better comments. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/408/c8e57e72bbd4b5c959838b9d6170258af9bbbff5..d32a358db475df2690ff956b608dd4fecc57099d)*

nikomatsakis (approved):
This looks good, I'd like to see it in action before making more suggestions, so I'm going to merge

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/408/c8e57e72bbd4b5c959838b9d6170258af9bbbff5..ff3c8ec32c5fa1eb6567329474e4e3e9523a75a6)*

PR #409, opened 2026-06-29
WIP: Rework inference

tiif:
## What does this PR do?

Refactor how we handle substitution. Don't review it yet, it's still WIP, push this so it is easier to read and fetch. 

## AI disclosure

* I did not use any AI tools


tiif:
This is mostly ok, there are a few questions left:

1. should we keep ``seq``? I initially thought we no longer need ``seq`` as we always apply the substitution in ``prove``, but there are some other places (``zip`` and ``for_all``) that still use ``seq`` in a way that couldn't be replaced with ``prove``. 
2. I think we should update the variable field in ``Env`` when calling ``add_substitution``? Do we have assumption such as every variable there is unique?

crates/formality-rust/src/prove/prove/prove/env.rs
```diff
@@ -355,6 +471,9 @@ impl Env {
             bias: self.bias,
             pending: vs.apply(&self.pending),
             allow_pending_outlives: self.allow_pending_outlives,
+            known_true: self.known_true,
+            // TODO: tiif thinks we need to update the substitution here, but tiif might be wrong
+            substitution: vs.apply(&self.substitution()),
```
tiif, line 476: Is this correct?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/409/c8e57e72bbd4b5c959838b9d6170258af9bbbff5..845edbb2cf2b6f05946dc5dd56d3cabe4806d2fe)*

crates/formality-rust/src/prove/prove/prove/prove_wc.rs
```diff
@@ -78,96 +75,96 @@ judgment_fn! {
...
 
         (
             (i in decls.neg_impl_decls(&trait_ref.trait_id))
             (let (env, subst) = env.existential_substitution(&i.binder))
             (let i = i.binder.instantiate_with(&subst).unwrap())
-            (prove(decls, env, assumptions, Wcs::all_eq(&trait_ref.parameters, &i.trait_ref.parameters)) => c)
-            (prove_after(decls, c, assumptions, &i.where_clause) => c)
+            (prove(decls, env, assumptions, Wcs::all_eq(&trait_ref.parameters, &i.trait_ref.parameters)) => env)
+            // TODO: check the correctness of this?
+            (prove(decls, env, assumptions,  &i.where_clause) => env)
```
tiif, line 102: It seems weird to have two proves, but this is basically extracting the previous logic. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/409/c8e57e72bbd4b5c959838b9d6170258af9bbbff5..845edbb2cf2b6f05946dc5dd56d3cabe4806d2fe)*

PR #430, opened 2026-07-09
Optional else branch

### LINKED ISSUE #429: Make `else` branch optional in `if` statements
## Description

Currently, `if` statements in formality require an `else` branch:

```rust
#[grammar(if $condition $then_block else $else_block)]
If {
    condition: Expr,
    then_block: Block,
    else_block: Block,
},
```

We should make the `else` optional — `if condition { ... }` without an else should be valid (with the else branch implicitly being `{}`/unit).

## Mentoring notes

**Where to start:**

Look at `crates/formality-rust/src/grammar/expr/mod.rs` around line 50.

**Options:**
1. Make `else_block` an `Option<Block>` and adjust the grammar annotation
2. Or keep two variants: `If` (with else) and `IfNoElse` (without), where `IfNoElse` desugars to `If` with an empty block

Option 1 is cleaner. You'll need to check if the `#[grammar(...)]` macro supports optional fields — if not, you may need a custom parse impl.

**Things to consider:**
- An `if` without `else` has type `()` — the then-branch must also have type `()` (or be a statement context)
- The borrow checker needs to handle the case where only one branch exists (the "else" path is implicitly "do nothing")
- Update the type checker and borrow checker to handle `None` for the else block

**Steps:**
1. Change the grammar to make else optional
2. Update type checking — if no else, the if-expression has type `()`
3. Update borrow checker — at join point, the "else" path has the same state as before the `if`
4. Add tests: `if` with and without else

gmemuriuki:
## What does this PR do?

Closes #429

Change the behavior of an if statement requiring an else statement by making the else statement optional through changing the grammar definition for the If statement. 
Create a small implementation for the Block in order to create an empty block for desugaring the empty else block.

Add a else_block variable in the judgement_fn! for the if rule to ensure that the block is not an option but either an empty block or a block in nll. Same thing to the codegen judgement_fn!

Change the macro since with the old macro, after else was consumed the parser tried Option<Block> which can succeed with either Some(block) or None, so `if c {} else {}` had two valid parses ( the {} could be a separate block statement totally unrelated with else). Now the block is mandatory when else is seen. 

I've also changed the formality core parser regarding the `expect_keyword()` and `expect_keyword_in()` functions since they 'ate' the identifier even when not a keyword such that `if {} let z ..` broke. So fix to restore position on mismatch. I'm not sure if that is right. Is it?

I also did a textual `Option<T>` detection in macro, Is it ok?

## AI disclosure


crates/formality-core/src/parse/parser.rs
```diff
@@ -596,28 +596,37 @@ where
...
-                skip_whitespace(text0),
-                format!("expected `{}`", expected),
-            )),
+            _ => {
+                self.current_text = text0;
+                Err(ParseError::at(
+                    skip_whitespace(text0),
+                    format!("expected `{}`", expected),
+                ))
+            }
```
tiif, line 612: I don't understand the change here 🤔 

why are we adding ``self.current_text = text0;`` when we already have ``let text0 = self.current_text;`` on top? isn't ``self.current_text = text0;`` just a no-op?

tests/mir_typeck.rs
tiif, line 1: It's ok to not have this test in ``mir_typeck``, we already have plenty of optional else branch test in ``tests/borrowck``.

tiif (commented):
Thanks! I am not comfortable with reviewing the parsing part, @nikomatsakis's input will be appreciated. For the rest, it looks good to me, I just have two questions. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/430/58e4bf41285eeba108e3ecf1db2ab583c1533523..b4054a628eb4d45d48746c8d8ff4806ca4b40cd3)*

tests/mir_typeck.rs
gme-muriuki: That's fine, I'll remove it.

crates/formality-core/src/parse/parser.rs
gme-muriuki: `identifier_like_string()` is mutating `self.current_text` -- it moves the cursor past the identifier before we compare it to expected (which happens in the match). So the `self.current_text = text0` is to return it to its original state when the match fails.

(I thought to include it so that if it is a desired behavior we do incorporate it if not, I can rework and get it out.)

gmemuriuki:
Thank you @tiif

crates/formality-rust/src/to_rust/syntax.rs
```diff
@@ -875,7 +875,7 @@ pub enum Stmt {
     If {
         condition: Expr,
         then_block: Block,
-        else_block: Block,
+        else_block: Option<Block>,
```
nikomatsakis:
```suggestion
        else_block: ElseBlock,
```

You know, we could do this:

```rust
#[term]
struct ElseBlock {
    block: Block,
}

impl Default for ElseBlock {
    fn default() -> Self { Block::empty() }
}
```

and then the rest of the code basically just do `else_block.block` and doesn't have to know if it was present or not.

nikomatsakis (changes requested):
I'm of two minds. On the one hand, I think these simplifications should take place elsewhere in a-mir-formality (i.e., be handled by an explicit desugaring step).

On the *other* hand, I think it's convenient, and fairly easy to support this kind of case,so maybe we should keep it. But what do you think about this simpler implementation? I'd be happier if it was "hidden" from the IR.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/430/58e4bf41285eeba108e3ecf1db2ab583c1533523..f721729fa37e7782a8f3abed340bc7c6dfae2d65)*

crates/formality-rust/src/to_rust/syntax.rs
gme-muriuki: That looks good. Shouldn't the struct elseblock be something like this:
```rust
#[term(else $block)]
struct ElseBlock {
  block: Block,
}
```

nikomatsakis (approved):
r+, this looks great

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/430/2d1941e68e2f03ff2b49e2568232ce29b6a5021a..d6c7db5793dabf90e3c41010850f35f416ab8402)*

gmemuriuki:
Thanks

PR #434, opened 2026-07-13
fix: Accept Rust-style integer literal syntax like `17_u32`

### LINKED ISSUE #427: Improve integer literal syntax (`0_u32` instead of `0 _ u32`)
## Description

Currently, integer literals in formality tests must be written as `0 _ u32` (with spaces around the underscore). This is awkward — we should support `0_u32` to match Rust's actual literal syntax.

The grammar annotation is in `crates/formality-rust/src/grammar/expr/mod.rs`:

```rust
/// `42_u32`
///
/// A scalar literal.
#[grammar($value _ $ty)]
Literal { value: usize, ty: ScalarId },
```

The comment already says `42_u32` but parsing actually requires spaces: `42 _ u32`.

## Mentoring notes

**Where to start:**

The issue is in how the `#[grammar($value _ $ty)]` macro expands — it treats `_` as a keyword token that requires whitespace around it. You likely need to write a custom parse impl (similar to how `&mut` types have custom parsing in `parse_impls.rs`) that parses `<digits>_<type-suffix>` as a single token without requiring spaces.

Look at:
- `crates/formality-rust/src/grammar/expr/mod.rs:127` — the current grammar annotation
- `crates/formality-core/` — how `#[grammar(...)]` works and how custom parse impls override it

**Steps:**
1. Write a custom parse impl for `Expr::Literal` that accepts both `0_u32` and `0 _ u32`
2. Update the debug/display impl if needed
3. Update tests to use the nicer syntax

gmemuriuki:
## What does this PR do?
Extract `Literal` into its own struct with a handwritten parser that accepts both `22_u32` and `22 _ u32`.

Update all the tests to use the nicer syntax for integer literal.

Closes #427

I've extracted the Literal enum variant into its own struct with term and customize macro while the Literal enum tuple has a cast macro.

Handwrite a parser that implements `CoreParse<FormalityLang>` trait, with the parse function implementation allowing both `22_u32` and `22 _ u32`. Use Parser::single_variant() a shorthand to create a parser for a  nonterminal with single variant.
The implementation rejects `Bool` ScalarId. e.g `22_bool` or `22 _ bool`.

Question: What did I not do?

## AI disclosure


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -417,7 +417,7 @@ judgment_fn! {
 
         (
             ------------------------------------------------------------ ("literal")
-            (borrow_check_expr(_env, _assumptions, state, Expr::Literal { value: _, ty }, _places_live_on_exit) => (ty, state))
+            (borrow_check_expr(_env, _assumptions, state, Expr::Literal(Literal { value: _, ty }), _places_live_on_exit) => (ty, state))
```
nikomatsakis:
```suggestion
            (borrow_check_expr(_env, _assumptions, state, Literal { value: _, ty }, _places_live_on_exit) => (ty, state))
```

we should be able to use upcasts so we can write the rule this way

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/434/0f2ec7f531c2fecdf49b690eff7e2e6e868a35b8..cb2788c15b6f210478aabd9adabfbedf857585ec)*

crates/formality-rust/src/codegen/mod.rs
```diff
@@ -231,7 +231,7 @@ judgment_fn! {
         (
             (let code = cfn.fresh_code_block())
             ---- ("literal")
-            (codegen_expr_into(global, cfn, scope, target, Expr::Literal { value, ty }) => (
+            (codegen_expr_into(global, cfn, scope, target, Expr::Literal( Literal{ value, ty })) => (
```
nikomatsakis: similarly here, we should be able to just write `Literal`

nikomatsakis (changes requested):
Let's fix the upcast issue, but otherwise, this looks great!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/434/0f2ec7f531c2fecdf49b690eff7e2e6e868a35b8..cb2788c15b6f210478aabd9adabfbedf857585ec)*

gmemuriuki:
Thanks for the review. I'm working on that and will have an updated PR soon

nikomatsakis (approved):

PR #435, opened 2026-07-15
Reject move from struct that implements drop

### LINKED ISSUE #422: Drop: no moving from a struct that implements Drop
## Description

The borrow checker must prevent moving individual fields out of a value whose type implements `Drop`. If you could move a field out, the `Drop` impl would run on a partially-moved value, which is unsound.

```rust
struct Foo { x: String, y: String }
impl Drop for Foo { ... }

fn bar(f: Foo) {
    let s = f.x;  // ERROR: cannot move out of `f.x` because `Foo` implements `Drop`
}
```

## Mentoring notes

**Where to start:**

This builds on the initialization/moves tracking (#296). Once we track moves at the place level, add an additional check: when moving from `place.field`, check whether `place`'s type implements `Drop`. If so, reject.

Write a judgment like `type_has_drop(decls, ty) => ()` that checks whether a type has a Drop impl. Because of the "always applicable" rules (enforced by the sibling issue), this is simpler than a full trait check — it's a binary predicate. Just search the declared impls for one whose trait is `Drop` and whose self type unifies with the struct. If such an impl exists, we know it matches (always-applicable guarantees it).

**Steps:**
1. Implement or wait for #296 (initialization/moves tracking)
2. Write the `type_has_drop` judgment — scan impl declarations for a `Drop` impl whose self type matches
3. When processing a move of `x.field`, look up the type of `x`
4. Call `type_has_drop` — if it succeeds, emit an error

kirloo:
## What does this PR do?

Resolves #422 

I reused the logic from `prove_ty_is_copy` for a function `prove_ty_is_drop` and added a predicate that `prove_ty_is_drop` should fail for the `TypedPlaceExpressionData::Field` branch of `prove_place_is_movable`.

`prove_ty_is_copy` and `prove_ty_is_drop` are nearly the same function, so maybe I could have refactored into a function like "prove_ty_is_trait".

## AI disclosure


nikomatsakis (changes requested):
[See comments on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/402470-t-types.2Fformality/topic/Office.20hours/near/610807398). Let's restructure this like so:

* extend `TypedPlaceExpressionData::Field` with the `AdtId` and variant info
* write a helper on `Program` to search for an impl of `Drop` for a given adt, taking advantage that we don't care about where-clauses etc
* then in the rule we can just check like `(if let None = program.find_drop_impl(adt_id))`

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/435/0f2ec7f531c2fecdf49b690eff7e2e6e868a35b8..416ce7f68b395ed70c44d8554bcb361dfa051c0d)*

kirloo:
@nikomatsakis I'm curious about what you think now. I had to split the `TypedPlaceExpressionData::Field` variant into two variants, `Field` and `TupleField`, since tuples don't have an `AdtId`.

nikomatsakis (approved):
This looks great @kirloo !

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/435/0509b0ad62cb5366b17d2da2b84772cce074c9b5..d51deea214ad397fad52be6db9f3783cfdb82c4d)*

PR #431, opened 2026-07-10
Flatten prove::prove::prove module nesting (#418)

Mslady-jojo:
Problem

The `prove` module in `formality-rust` was nested three levels deep (`prove::prove::prove::*`), with `prove.rs` declaring a submodule `prove` (`prove/prove.rs`), which itself declared another submodule `prove` (`prove/prove/prove.rs`) containing the actual proving logic (`env`, `prove_eq`, `prove_wf`, `combinators`, etc). This meant every import path across the codebase had to spell out `crate::prove::prove::prove::X` or `crate::prove::prove::X` just to reach ordinary sibling items, with no corresponding conceptual separation to justify the extra layers.

What changed:

- Moved all files from `prove/prove/prove/*` and `prove/prove/*` up into `prove/*` directly (`git mv`, so history is preserved).
- Merged the two wrapper files (`prove/prove.rs` and `prove/prove/prove.rs`) directly into the top-level `prove.rs`, combining their `mod` declarations and `pub use` re-exports (de-duplicating overlapping re-exports like `Constrained`/`Constraints` and the negation functions, which were previously re-exported from two places).
- Updated all 48 call sites across the codebase (in `check/`, `codegen/`, `to_rust/`, `grammar/`, and within `prove/` itself) from `crate::prove::prove::X` / `crate::prove::prove::prove::X` to the flat `crate::prove::X`.
- Regenerated `expect_test` snapshots in `tests/basic_tests.rs` where the expected output embedded the old (now-changed) source file paths in panic/error messages. No logic or assertion values changed only the file path strings.
`cargo clippy --all-targets` : same pre-existing warning count as `main`; no new warnings introduced, and none touch files modified in this PR. Left untouched to keep this PR scoped to the module restructuring only.
This is a pure structural refactor no behavior changes.

PR #436, opened 2026-07-16
Enforce that Drop impls are in the ADT's defining crate

### LINKED ISSUE #421: Drop: enforce "always applicable" rules
## Description

We should enforce the rules that `Drop` can only be explicitly implemented on structs or enums in the same crate, and only when it is "always applicable."

"Always applicable" means: the impl's bounds must be implied by the struct's own bounds. For example:

```rust
struct Foo<T: Debug> { }

impl<T: Debug + Clone> Drop for Foo<T> { }  // ERROR: `Clone` is not always applicable
```

This is required because dropping happens implicitly — the compiler can't insert conditional drop calls based on whether extra bounds hold.

## Mentoring notes

**The core check:** Given a struct and its Drop impl, we want to prove:

```
for all struct parameters PS0...PSn:
    if struct where-clauses hold:
        exists impl parameters PI0...PIn:
            Struct<PS0..PSn> = impl Self type &&
            impl where-clauses are provable
```

In other words: for any valid instantiation of the struct, we can find an instantiation of the impl that matches. This ensures the Drop impl is "always applicable" — any time you can construct the struct, you can drop it.

**Where to start:**

1. Look at how trait impls are checked: `crates/formality-rust/src/check/impls.rs`
2. You'll need to assemble the proof goal above and prove it using the existing `prove` infrastructure
3. You'll also need to verify `Drop` is only implemented for types defined in the same crate
4. Look at rustc's implementation for reference: `rustc_hir_analysis::check::dropck`

**Test cases to add:**
- `Drop` impl with matching bounds → OK
- `Drop` impl with extra bounds → ERROR
- `Drop` impl for a type from another crate → ERROR

System625:
## What does this PR do?

Most of #421 already landed in #369 (commit 3a799923): `check_drop_impl_always_applicable` proves the "always applicable" goal and rejects non-struct/enum self types, with tests in `tests/drop.rs`. This PR adds the remaining piece: a `Drop` impl must live in the crate that defines the ADT. Previously that case was only rejected indirectly by the coherence orphan check, with a generic `is_local` error. Now it fails fast with a Drop-specific message: ``` `Drop` may only be implemented for types defined in the current crate: `MyStruct` is defined in crate `a` but the impl is in crate `b` ``` 
Also adds test coverage the issue asked for that was missing: enums (OK and extra-where-clause error) and a cross-crate OK case.
Closes #421. 

So basically, "any valid instantiation" just means: take the struct's generics/where-clauses as given (like `T: Clone` for `MyStruct<T>`), and treat `T` as some opaque type that only satisfies those bounds, nothing more, nothing less. We're not allowed to assume extra stuff about `T`.
For `prove_goal`, the checker only has the ADT's own where-clauses to work with. It asks the trait solver "can this type prove `Drop`?" by unifying the self type with `MyStruct<T>` and trying to prove it. If the impl needs something the struct's bounds don't already guarantee (say `T: Debug`), the proof fails, so the impl gets rejected. That's what "always applicable" means: the impl has to work for every valid `T`, not just some.
The actual new part in this PR: before running the checker, we figure out which crate owns the struct/enum, and require the `Drop` impl to live in that same crate. Previously this was only caught by the general orphan rule, and errors just talked about locality. Now we can point at coherence directly and name both crates.

## AI disclosure

### DIFF
--- a/crates/formality-rust/src/check/impls.rs
+++ b/crates/formality-rust/src/check/impls.rs
@@ -281,25 +281,45 @@ fn drop_impl_adt_id(trait_impl: &TraitImpl) -> Fallible<AdtId> {
     Ok(adt_id.clone())
 }
 
+/// Check that a `Drop` impl declared in crate `crate_id` is for an ADT defined in that
+/// same crate.
+fn check_drop_impl_in_defining_crate(
+    program: &Program,
+    adt_id: &AdtId,
+    crate_id: &CrateId,
+) -> Fallible<ProofTree> {
+    let defining_crate = program.program().crate_defining_adt(adt_id)?;
+    if defining_crate != crate_id {
+        bail!(
+            "`Drop` may only be implemented for types defined in the current crate: \
+            `{adt_id:?}` is defined in crate `{defining_crate:?}` but the impl is in crate `{crate_id:?}`"
+        );
+    }
+    Ok(ProofTree::leaf("check_drop_impl_in_defining_crate"))
+}
+
 judgment_fn! {
     /// Check that a `Drop` impl is "always applicable": for any instance of the ADT
     /// (with its where-clauses satisfied), the Drop impl must apply.
     pub(super) fn check_drop_impl_always_applicable(
         program: Program,
         trait_impl: TraitImpl,
+        crate_id: CrateId,
     ) => () {
-        debug(program, trait_impl)
+        debug(program, trait_impl, crate_id)
 
         (
             (if **trait_impl.trait_id() != *"Drop")!
             ---- ("not a Drop impl")
-            (check_drop_impl_always_applicable(program, trait_impl) => ())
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):

System625:
@nikomatsakis the merge queue kicked this back out on a failing check. Root cause was a stale expect-test snapshot, not a real regression.

The drop_impl_enum_extra_where_clause test I added had its snapshot hard-coding the old prove/prove/prove/prove_via.rs:9 path. This branch was cut before #431 flattened the prove module, so my own CI passed against the old main, but the merge queue re-ran it against current main (where the path is now prove/prove_via.rs:8) and the snapshot no longer matched.

So i pusehd a fix by rebasing onto current main and regenerated the snapshot. Net change is a single line (:9 → :8 on that one test). it should be good to re-queue whenever you get a chance.

PR #440, opened 2026-07-26
Raw pointers are not fundamental types

bit-aloo:
## What does this PR do?

I am currently investigating https://github.com/rust-lang/a-mir-formality/issues/387 In Formaility, we were incorrectly treating `RawPointer` as a fundamental type during coherence checking. This PR fixes that and also adds a regression test.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/prove/is_local.rs
+++ b/crates/formality-rust/src/prove/is_local.rs
@@ -327,13 +327,13 @@ fn is_fundamental(_decls: &Program, name: &RigidName) -> bool {
         RigidName::AdtId(_) => false, // FIXME(#222)
 
         RigidName::Ref(_) => true,
-        RigidName::Raw(_) => true,
 
         RigidName::Never => false,
 
         RigidName::ScalarId(_)
         | RigidName::Tuple(_)
         | RigidName::FnPtr(_)
-        | RigidName::FnDef(_) => false,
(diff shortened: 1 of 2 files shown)


nikomatsakis (approved):
Nice!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/440/c490389679ab0d790cfad83f20dccae46537589e..61bf6d8cfd7067456550279327e6aff9453631a8)*

nikomatsakis:
Does this actually *fix* https://github.com/rust-lang/a-mir-formality/issues/387, @bit-aloo? It seems like no, right?

bit-aloo:
> Does this actually _fix_ #387, @bit-aloo? It seems like no, right?

Nope, 

I am still looking into it. From what I have found so far, Formality and rustc model coherence a bit differently. During coherence checking, rustc may leave a projection unnormalized if its trait reference isn’t knowable, while formality normalizes it in this case. That seems to be the reason for the discrepancy in #387.

I initially tried a small workaround to make the test pass, but it was more of a hack than the right solution. I am now digging deeper into rustc’s coherence logic, helping around next solver coherence side of changes, before trying to fix it here properly. If you have any suggestions, that would be really awesome.

nikomatsakis (approved):

PR #442, opened 2026-07-27
Make ref expr with lt order first

### LINKED ISSUE #441: Inconsistent mutable reference syntax
The example below is a pass test. As shown in the example below, we use ``&'a mut`` syntax for input ty and LHS of an assign statement, but ``&mut 'c `` for RHS. It will be nice if we just stick to one of the syntaxes. 

```rust
#[test]
fn reborrow_check() {
        struct Map { }

        fn foo<'a>(m: &'a mut Map) -> () {
            exists<'b, 'c> {
                let y: &'b mut Map = &mut 'c *m;
            }
        }
    }])
    .skip_execute()
    .ok()
}
```

This example was discovered while working with @dingxiangfei2009

bit-aloo:
## What does this PR do?

closes #441

https://github.com/rust-lang/a-mir-formality/issues/441#issuecomment-5089512291

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/check/borrow_check/nll.rs
+++ b/crates/formality-rust/src/check/borrow_check/nll.rs
@@ -489,7 +489,7 @@ judgment_fn! {
             // Reborrow constraints change the outlives set without going
             // through `TypeckEnv::prove_judgment`, so run the universal
             // outlives check here as well; otherwise a constraint between
-            // universal regions (e.g. `&mut 'b *p` with `p: &mut 'a T`)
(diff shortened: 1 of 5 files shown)


nikomatsakis (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/442/c490389679ab0d790cfad83f20dccae46537589e..94aaea1b73135f3f80b493e040c97bacf3ceeba3)*

Issue #209, opened 2025-12-17
Extend type/borrow-check to guarantee values are returned


nikomatsakis:
Functions can return without initializing their return place, which should be a type/borrow check error.

Currently the check for return place initialization was removed to enable the pending_outlives refactoring (threading outlives constraints as parameters instead of mutating TypeckEnv).

The test `test_uninitialised_return_type` demonstrates a function that returns without initializing the return place - this should fail but currently passes.

This validation should be restored, potentially by:
1. Adding return place initialization tracking to the borrow checker
2. Making it part of the type checking phase in a different way
3. Adding it as a separate validation pass

tiif:
Hi @Sandijigs, are you still interested in working on this issue?

PR #444, opened 2026-07-28
fix: reject missing return values

### LINKED ISSUE #392: Make sure we errror for type mismatch when nothing is returned
We should make sure we throw an error for the test below. The function signature says ``u32`` should be returned, but nothing is returned in the function body. 
```rust
#[test]
fn i_am_a_test_that_should_be_renamed_later() {
        fn foo() -> u32 {
        }
    }])
    .ok()
}
```

gmemuriuki:
## What does this PR do?
closes #392

Adds return validation that rejects non-unit functions whose bodies can reach end without them returning a value.

This PR introduces a control flow summary containing:

- Whether execution can fall through to the next statement.
- labels targeted by `break`.
- labels targeted by `continue`

The return checker sequentially analyzes statements and blocks:

- Sequential statements are composed using `then`.
- Alternative branches are combined using `join`.
- ...

For functions returning `()`, falling through remains valid. For non-unit functions, validation succeeds only when the function body cannot reach its end. This includes functions where every reachable path returns, as well as loops that cannot exit.

Q: Did this implementation halfway solve what this issue needs? #209

## AI disclosure

* Other: I used an AI to help reason through the control-flow design and test. I wrote the implementation.


crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -261,25 +262,30 @@ impl FlowState {
...
         }
     }
 
     pub fn with_break(&self, label: &LabelId) -> Self {
         let mut this = self.clone();
-        this.breaks.insert(LabeledFlowState {
-            label: label.clone(),
-            state: this.current.clone(),
-        });
+        if !this.diverged {
```
nikomatsakis: This is interesting. We could probably make a good borrow checker test here too, something like

```rust
let mut a = 1;
let mut b = 2;

let mut p = &a;
loop {
    break;
    p = &b;
    break;
}

b += 1;
use(p);
```

I think that before your changes, we would reject this -- but not after.

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -261,25 +262,30 @@ impl FlowState {
...
         this
     }
 
     pub fn with_continue(&self, label: &LabelId) -> Self {
         let mut this = self.clone();
-        this.continues.insert(LabeledFlowState {
-            label: label.clone(),
-            state: this.current.clone(),
-        });
+        if !this.diverged {
```
nikomatsakis: we can make a similar test here, I didn't see any test for this

crates/formality-rust/src/check/borrow_check/flow_state.rs
```diff
@@ -501,6 +508,8 @@ impl FlowState {
             .into_iter()
             .partition(|lfs| Some(&lfs.label) == scope_label.as_ref());
         let mut successor = current;
+        // A break targeting this scope is live control arriving after it.
+        let diverged = diverged && this_label.is_empty();
```
nikomatsakis: This condition doesn't seem quite right to me. For example

```rust
fn foo() {
    let mut x = 1;
    let mut y = 1;
    let mut p = &x;
    'b: {
        'a: {
            break 'b;
        }
        // I think we would conclude "diverged = false" here
        p = &y;
    }
    y += 1; // ...and hence incorrectly report an error here
    use(p);
}
```

nikomatsakis (changes requested):
Take a look @gme-muriuki at the examples I raised

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/444/c490389679ab0d790cfad83f20dccae46537589e..01a09b588443d3ef6a7f0d2306f0a950da4730f1)*

crates/formality-rust/src/check/borrow_check/flow_state.rs
gme-muriuki: Adding it rn 👍

crates/formality-rust/src/check/borrow_check/flow_state.rs
gme-muriuki: I'm not sure if this is a good thing or bad that it has changed.  But I'll rework the whole approach and preserve the desired behavior.

crates/formality-rust/src/check/borrow_check/flow_state.rs
gme-muriuki: This made me rethink the entire approach. Take a look and tell me whether it does still fail for this case. Happy to keep iterating.

Issue #417, opened 2026-07-08
Consolidate `Wc`, `WhereClause`, `Predicate`, and `Relation`


nikomatsakis:
## Description

We currently have too many overlapping concepts for representing "things to prove." We should align on a smaller set.

**Current state:**
- `WhereClause` — syntactic representation written in Rust source (`crates/formality-rust/src/grammar/traits_and_impls.rs:128`)
- `Wc` — internal logical goal, wraps `Predicate` or `Relation` plus `ForAll`/`Implies` (`crates/formality-rust/src/grammar/wc.rs:149`)
- `Predicate` — atomic provable facts like `IsImplemented`, `AliasEq` (`crates/formality-rust/src/grammar/formulas.rs:18`)
- `Relation` — built-in goals like `Equals`, `Sub`, `Outlives`, `WellFormed` (`crates/formality-rust/src/grammar/formulas.rs:129`)

**Proposed structure:**
- `WhereClause` — stays as the syntactic level, lowered to `Predicate`
- `Predicate` — the single internal representation, combining today's `Wc`/`Predicate`/`Relation`

## Mentoring notes

**Where to start:**

Look at how `WhereClause` is lowered. The `ToWcs` trait in `crates/formality-rust/src/prove.rs` is the bridge:

```rust
pub trait ToWcs {
    fn to_wcs(&self) -> Wcs;
}
```

And `WhereClause` implements it to convert syntax into internal goals.

**Steps:**
1. Map out all consumers of `Wc`, `Predicate`, and `Relation` (grep for uses)
2. Sketch a unified `Predicate` enum that has all the variants from both `Predicate` and `Relation`, plus the `ForAll`/`Implies` connectives currently in `Wc`
3. Migrate incrementally — perhaps start by merging `Relation` into `Predicate` since they're both "atomic things to prove"
4. Then collapse `Wc` into the new `Predicate`
5. Assess how this impacts the judgment rules (the rules dispatch on these types)

**Risk:** This touches a lot of files. Plan to do it in multiple PRs (e.g., merge Relation into Predicate first, then collapse Wc).

PR #445, opened 2026-08-03
Remove the leftover *Data migration aliases

System625:
## What does this PR do?

Removes the six `pub type FooData = Foo` migration aliases left over from the #318 flatten series (#321, #338, #353) and renames their call sites: `WcData`, `WhereClauseData`, `WhereBoundData`, `TyData`, `LtData`, `ConstData`. Also updates one doc example in `formality-macros` that still named `TyData`. This is a pure rename with nothing left behind
The diff change is much because there are `expect!` snapshots recording that `prove_normalize_via` moved from line 56 to 54 once its import collapsed to one line; normalizing those numbers out of both sides of the diff makes the two sides identical.
Groundwork for #417.

## AI disclosure


nikomatsakis (approved):
Oh nice, I've been wanting to do this

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/445/c490389679ab0d790cfad83f20dccae46537589e..87f3d5b27e77de93990db289b2da9d9203e010f0)*

PR #446, opened 2026-08-04
Refactor borrowck tests to always test all three modes

jackh726:
## What does this PR do?

- Test for all three borrowck modes for borrowck tests
- Don't duplicate errors
- Ensure that unlocked is more permissive than alpha, which is more permissive than nll

## AI disclosure

### DIFF
--- a/src/test_util.rs
+++ b/src/test_util.rs
@@ -1,7 +1,8 @@
 use expect_test::Expect;
 use formality_core::test_util::AnyhowResultTestExt;
+use formality_rust::grammar::FeatureGateName;
 
-use crate::{run_rustc, test_program_ok};
+use crate::{run_rustc, test_program_ok, test_program_ok_with_feature_gates};
 
 /// Stringify a list of crate declarations and wrap them in the `[ ... ]`
 /// brackets that the Crates grammar expects.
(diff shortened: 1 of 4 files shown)


src/test_util.rs
```diff
@@ -106,6 +114,145 @@ impl FormalityTest {
...
+    }
+}
+
+/// A borrow-check test: one program, run under all three borrowck modes.
+///
+/// The program is given as crate *items* (see the `items!` macro); the crate
+/// wrapper and feature gate are supplied per mode. Every mode that rejects the
+/// program must reject it with the *same* error, so a test records one expected
+/// error rather than three.
+pub struct BorrowckTest {
```
nikomatsakis: Is there a reason you made this as an alternative to `FormalityTest` versus adding it as an option (e.g., `FormalityTest::new().borrow_check().run`)?

It seems like merging it into `FormalityTest` would help us detect conflicts as we add other modes (e.g., comparing against rustc).



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/446/2d1941e68e2f03ff2b49e2568232ce29b6a5021a..c36743b36035b71a3d1319bf7eab8be8c071a3d6)*

nikomatsakis (changes requested):

nikomatsakis (approved):

Issue #419, opened 2026-07-08
Add `mut` declarations and mutable-place checks


nikomatsakis:
## Description

Currently, formality doesn't distinguish between `let x` and `let mut x`. All variables are implicitly mutable. We need to:

1. Add `mut` as an optional annotation on `let` declarations in the grammar
2. Add "mut place" checks into the borrow checker rules — when performing a mutable access (assignment, `&mut` borrow), verify that the target place is mutable:
   - A local variable and the data it owns is `mut` if declared `let mut x`
   - When passing through a deref, deref of `&` is not mut, but deref of `&mut` is

## Mentoring notes

**Where to start:**

1. **Grammar:** Look at `crates/formality-rust/src/grammar/expr/mod.rs` for how local variable declarations work. You'll need to add a `Mutability` field (or just a bool) to the `let` statement representation.

2. **Borrow checker:** Look at `crates/formality-rust/src/check/borrow_check/` — when an assignment or `&mut` borrow is processed, add a check that the target local was declared `mut`.

3. **Existing tests:** The existing borrow check tests in `tests/borrowck.rs` will need updating — they'll need `mut` annotations where variables are mutated.

**Steps:**
1. Add `Mutability` enum (`Mut`/`Not`) to the grammar for local declarations
2. Thread it through parsing/lowering
3. Create a judgment `prove_place_is_mut(place) => ()` with rules:
   - A local variable `x` is mut if declared `let mut x`
   - A deref of `&mut T` is mut
   - A deref of `&T` is NOT mut
   - A field access `place.field` is mut if `place` is mut
4. In the borrow checker, when you see a write to a place or a `&mut` borrow of a place, call `prove_place_is_mut`
5. Add test cases: mutable access to non-mut variable (should error), mutable access to mut variable (should pass), write through `&mut` (should pass), write through `&` (should error)
6. Update all existing tests that mutate variables to use `mut`

PR #447, opened 2026-08-07
Check for duplicate trait impl item

bit-aloo:
## What does this PR do?

We didn't had any check for duplicate trait impl item. This PR adds that. I am currently working through https://github.com/rust-lang/a-mir-formality/issues/419, and going across all checks to understand better. I believe there might be more such cases, I will try to spot more.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/check/impls.rs
+++ b/crates/formality-rust/src/check/impls.rs
@@ -1,4 +1,5 @@
 use anyhow::bail;
+use formality_core::Set;
 
 use crate::grammar::{
     AdtId, AssociatedTy, AssociatedTyBoundData, AssociatedTyValue, AssociatedTyValueBoundData,
@@ -31,6 +32,7 @@ judgment_fn! {
             (let TraitBoundData { where_clauses: _, trait_items } = trait_decl.binder.instantiate_with(&trait_ref.parameters)?)
             (check_safety_matches(&trait_decl, &trait_impl) => ())
 
+            (check_duplicate_impl_items(impl_items) => ())
             (for_all(impl_item in impl_items)
                 (check_trait_impl_item(program, env, where_clauses, trait_items, impl_item, crate_id) => ()))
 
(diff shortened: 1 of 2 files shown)


crates/formality-rust/src/check/impls.rs
```diff
@@ -123,6 +125,27 @@ fn check_all_required_items_present(
...
 
+fn check_duplicate_impl_items(impl_items: &[ImplItem]) -> Fallible<ProofTree> {
+    let mut functions = Set::new();
+    let mut types = Set::new();
+
+    for impl_item in impl_items {
+        match impl_item {
+            ImplItem::AssociatedTyValue(AssociatedTyValue { id, .. }) => {
+                if !types.insert(id) {
+                    bail!("Assoc ty {id:?} defined multiple times");
```
tiif: nit:

```suggestion
bail!("assoc ty {id:?} is defined multiple times");
```

crates/formality-rust/src/check/impls.rs
```diff
@@ -123,6 +125,27 @@ fn check_all_required_items_present(
...
+                }
+            }
+            ImplItem::Fn(fun) => {
+                if !functions.insert(&fun.id) {
+                    bail!("Function item {:?} defined multiple times", &fun.id);
```
tiif: nit: 

```suggestion
bail!("function item {:?} is defined multiple times", &fun.id);
```

tiif (changes requested):
Thanks! Other than the nits, the rest are fine to me. 

(the fmt in my suggestion is off, you will need to run cargo fmt after applying the change)
*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/447/42973e92c9d97869ba107bb8981f8996eda998a0..9a84ab7191011006533f4c89e7d1172d8b61faf9)*

crates/formality-rust/src/check/impls.rs
bit-aloo: Done, thanks.

crates/formality-rust/src/check/impls.rs
bit-aloo: Done thanks.

tiif (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/447/192ec929a1f9ca280e913da7d43dcffd06a9dff2..46d9387f9d48fcf98e44d7cf9d972fc0d0b17cbe)*

PR #448, opened 2026-08-07
Gate access_kind_for_place_use AccessKind::Move rule on !prove_ty_is_copy

jackh726:
## What does this PR do?

See the two added tests: the AccessKind::Move rule was being used even when ty is copy.

Question: as @nikomatsakis pointed out to me, it's a bit surprising that the `Copy` rule doesn't result in a successful program despite the `Move` rule being selected (i.e. we should try *both*).

As for to_rust, use `status` for cargo fmt, not `spawn` (the former is blocking on the result, the latter is not).

## AI disclosure


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -947,8 +947,9 @@ judgment_fn! {
         )
 
         (
+            (if !prove_ty_is_copy(env, assumptions, state, &place.ty).is_proven())!
```
nikomatsakis, line 950: Yeah I feel like this should not be needed, I want to look into this more closely.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/448/42973e92c9d97869ba107bb8981f8996eda998a0..c04a5e0a8c093d73e9dbeb1630a7d288bd4f7a4b)*

nikomatsakis (changes requested):

PR #449, opened 2026-08-08
Check neg trait decl

bit-aloo:
## What does this PR do?

We didn't check if the trait for actually present during neg trait impl check, this PR adds that check

Adds the check for trait decls

## AI disclosure

* I did not use any AI tools


bit-aloo:
this seems super weird: https://github.com/rust-lang/a-mir-formality/actions/runs/31241271657/job/93062541590?pr=449, why formatting of the test changes here?

nikomatsakis (approved):
Ha! Good idea.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/449/42973e92c9d97869ba107bb8981f8996eda998a0..ef5f7be1d1bce585d61e6eb3e95588a5c7052922)*

PR #450, opened 2026-08-10
Merge Relation into Predicate

System625:
## What does this PR do?

Part of #417. Merges `Relation` into `Predicate` so there is a single atomic-goal type.

* `Predicate` gains `Equals` / `Sub` / `Outlives` / `WellFormed`, declared **before** the existing variants.
* `Relation`, its `debone()`, and the (unused) `Debone` trait + `debone_impl!` macro are deleted.
* Adds `Skeleton::is_relation()`, and drops the dead `Skeleton::IsInt` variant.
* `prove_wc`'s two byte-identical `"assumption - predicate"` / `"assumption - relation"` rules collapse into a single `"assumption"` rule.

`WhereClause` still lowers into `Wc`, and this PR does not touch that layer. Wc` survives, so the next PR in this series is a rename rather than a merge.

Also noticed while working in `grammar/formulas.rs`, and deliberately **not** fixed here: `Coinductive` is dead code (only its own definition plus a `BitAnd` impl, no other references anywhere in the workspace). Pre-existing and unrelated to this refactor, happy to remove it in a separate PR if you want it gone.

The merge is safe because `Skeleton`/`debone()` already lumped relations and predicates together before this PR, and nothing else in the grammar has a `Predicate` or `Relation` field, so folding the two enums into one doesn't touch anyone else's terms, but `prove_via` still needs to distinguish them, since `predicate-congruence-axiom` proves parameters equal while `relation-axiom` needs them literally identical (doing congruence on `Equals` would be circular), so now that both rules match `Wc::Predicate` instead of separate `Wc::Relation`/`Wc::Predicate` patterns, the split is handled by a `Skeleton::is_relation()` guard placed before the `!` in each rule, which keeps the same silent-skip behavior as before since `!` is the commit point for `push_rules!` failures; and finally, variant order still matters because `#[term]` derives `Ord` and `Wcs` is a sorted set so putting relations first inside `Predicate` reproduces the old `Relation | Predicate` ordering so set order and proof search order don't shift.

## AI disclosure

### DIFF
--- a/crates/formality-rust/src/prove/prove_via.rs
+++ b/crates/formality-rust/src/prove/prove_via.rs
@@ -26,6 +26,7 @@ judgment_fn! {
             (let (skel_c, parameters_c) = pred_1.debone())
             // `g` = "goal, the name for something that we are trying to prove.
             (let (skel_g, parameters_g) = pred_2.debone())
+            (if !skel_c.is_relation())
             (if skel_c == skel_g)!
             (prove(decls, env, assumptions, Wcs::all_eq(parameters_c, parameters_g)) => c)
             ----------------------------- ("predicate-congruence-axiom")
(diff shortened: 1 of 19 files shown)


crates/formality-rust/src/prove/prove_via.rs
```diff
@@ -26,6 +26,7 @@ judgment_fn! {
             (let (skel_c, parameters_c) = pred_1.debone())
             // `g` = "goal, the name for something that we are trying to prove.
             (let (skel_g, parameters_g) = pred_2.debone())
+            (if !skel_c.is_relation())
```
nikomatsakis, line 29: So I'm not super crazy about this part of the rules, but it's ok, let's merge the PR as is. I would like to eventually not have this distinction and instead to handle everything like using the `(if !skel_c.is_relation())`

nikomatsakis (approved):
So I think that this PR looks good. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/450/42973e92c9d97869ba107bb8981f8996eda998a0..7129bc2cb169155d9bf4724789b6f2799c5c3f31)*

PR #451, opened 2026-08-12
Wait for cargo fmt

bit-aloo:
## What does this PR do?

I have been seeing CI failures on another PR of mine: https://github.com/rust-lang/a-mir-formality/actions/runs/31603415405, which seem to be caused by deferred formatting. We used to spawn `fmt`, so it would run in the background, but now we wait for `fmt` to finish before doing anything else.

## AI disclosure

* I did not use any AI tools

### DIFF
--- a/crates/formality-rust/src/to_rust/mod.rs
+++ b/crates/formality-rust/src/to_rust/mod.rs
@@ -94,7 +94,7 @@ pub fn create_workspace(crates: &Crates, root_directory: &std::path::Path) -> Fa
 
     std::process::Command::new("cargo")
         .args(["fmt", "--all", "--manifest-path", location])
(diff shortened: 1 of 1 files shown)


lqd:
#448 also does this

bit-aloo:
> #448 also does this

Oh, I hadn't seen that, thanks! They might have run into the same issue too.

Not sure how long it'll take to merge, though. Maybe we can go with this for now so other prs don't hit the same issue. I am also fine with closing this. 🙂

lqd:
Yeah, with the holidays and all, that other PR will likely not be merged soon and can be rebased later, let's go with this one then!

lqd (approved):

PR #455, opened 2026-08-26
Add mut declarations and mutable-place checks

### LINKED ISSUE #419: Add `mut` declarations and mutable-place checks

bit-aloo:
## What does this PR do?

closes: #419 

took me sometime but i do understand borrow check a bit better now

## AI disclosure

* I did not use any AI tools


crates/formality-rust/src/check/borrow_check/nll.rs
```diff
@@ -1022,6 +1026,88 @@ judgment_fn! {
...
+    fn prove_place_is_assignable(
+        env: TypeckEnv,
+        assumptions: Wcs,
+        state: FlowState,
+        place: TypedPlaceExpr,
+    ) => () {
+        debug(place, state, assumptions, env)
+
+        (
+            (if !check_place_initialized(&state, &place.to_place_expression()))
```
xonx4l: I think `check_place_initialized` is also false after a move not just for `let` without an initializer.
```rust
let x: Datum = Datum { value: 0_u32 };  // not mut
let y: Datum = x;
x = Datum { value: 1_u32 };             // rustc:- E0384
```
Also can you add some `borrowck_err` tests? Nothing here checks the error cases.

PR #456, opened 2026-09-07
Add support for 128 bit integers

liolin:
## What does this PR do?

Add support for `i128` and `u128`.

Extend various `enum`'s and add boilerplate impls.

## AI disclosure

* I did not use any AI tools


crates/formality-core/src/parse.rs
tiif, line 1: Good catch! It's interesting how many parsing mistakes we had in this file.

tiif (changes requested):
Could you also add tests that use i128 and u128 in formality surface syntax and codegen to minirust (``tests/codegen.rs``)? It is to make sure the parsing actually works and it can successfully codegen. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/a-mir-formality/456/a207654fd83834ef075128f819f25d5d28d5122c..d852f99ad95e42f8197bb780e647e1297198aeb0)*

liolin:
I've added a test, and it actually failed. I had to adjust the type in the `Literal` struct to be `u128`. Could this potentially lead to problems?

tiif:
That's surprising. What error did you see after you added a test and before changing the ``Literal`` struct?

liolin:
Because `usize` is only 64 bit wide (on a 64 bit machine) the parser fails when trying to parse a number bigger than `usize::MAX`.

tiif:
Ah I see why now, what we have in the past is a bit iffy. If we use ``u128`` for literal, anything lesser than 0 will fail the parse. If we use ``i128`` for literal, then anything larger than ``i128::MAX`` will fail to parse (this is possible when we also support ``u128``, ``u128::MAX`` is larger than ``i128::MAX``). So using ``u128`` or ``i128`` as the type of ``value`` here won't make the issue go away. This is a preexisting issue that we didn't catch because we probably didn't test this back then. 

There are a few options:
1. Just support either unsigned int or signed int, and use ``u128``  or ``i128`` for ``Literal``
2. We split ``Literal`` into unsigned int version and signed int version, and have ``u128`` and ``i128`` as the type of the ``value`` respectively. 

@nikomatsakis I am curious about your thought on this, since you are most familiar with parsing infrastructure.

liolin:
If desired, I could try to implement the parse trait manually for `Literal`.

tiif:
@liolin Yeah, please feel free to try it :)

